PR #2, opened 2019-01-04
Added a system property, that enables the zip file cache for larger f…

Frontrider:
…iles.

If I understood it correctly, than line 253 recieves the uri of the jar file, which turns into a zipfilesystem under the hood.
I added a system property to set the proper environmental values. It's a system property, so that people who don't need it don't need to bother, and to make it plug into loom as is, without modifications.

CI servers can set the property when required.

### DIFF
--- a/src/main/java/net/fabricmc/tinyremapper/TinyRemapper.java
+++ b/src/main/java/net/fabricmc/tinyremapper/TinyRemapper.java
@@ -244,7 +244,13 @@ public class TinyRemapper {
 			}
 
 			if (fs == null) {
-				fs = FileSystems.newFileSystem(uri, Collections.emptyMap());
+                Map<String, Object> env = new HashMap<>();
+			    if(System.getProperty("tiny_use_zipcache","FALSE").equals("TRUE")){
+                    env.put("create", "true");
+                    env.put("useTempFile", Boolean.TRUE);
+                }
+
+				fs = FileSystems.newFileSystem(uri, env);
 				fsToClose.add(fs);
 			}
 


src/main/java/net/fabricmc/tinyremapper/TinyRemapper.java
```diff
@@ -244,7 +244,13 @@ public FileVisitResult visitFile(Path file, BasicFileAttributes attrs) throws IO
 			}
 
 			if (fs == null) {
-				fs = FileSystems.newFileSystem(uri, Collections.emptyMap());
+                Map<String, Object> env = new HashMap<>();
+			    if(System.getProperty("tiny_use_zipcache","FALSE").equals("TRUE")){
```
liach: Do people use lowercase or uppercase for values more often?

shadowfacts: Better yet, use lowercase and call toLowerCase on the property value. Furthermore, System.getProperty is null if the property isn't present, so this will crash with an NPE if the property isn't specified.

Frontrider: It can't do that, because I give a default value to be used when it's not present. I can set it to lowercase.

Although, it's only a partial fix, because it still runs out of memory, but it's no longer because of the zipfs.

sfPlayer1:
Can you elaborate on the benefits? Any measurements?

Frontrider:
It crashes on our ci server when it tries to load up the entire archive, as it runs out of memory. We can't give it more. Modmuss benchmarked it, the entire gradle build should use 1.1G, but 1.5 is not enough.

It fixes that issue, but then it starts to do the same as it makes byte buffers that are too large, So we might need a different fix.

src/main/java/net/fabricmc/tinyremapper/TinyRemapper.java
mikroskeem: My suggestion is to use `"tiny.useZipCache"`

Also - don't forget `Boolean.getBoolean()` :wink:

PR #3, opened 2019-01-24
use tiny-mappings-parser library

asiekierka:
Unfortunately, at this time it adds a noticeable (with the naked eye) performance hit to the mappings loading process - I want to look if I can get it down first.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -12,7 +12,7 @@ buildscript {
 }
 
 sourceCompatibility = 1.8
-version = '0.1.0'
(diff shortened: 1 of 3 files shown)


NebelNidas:
Can be closed now that #110 is merged. Although we might want to move the methods introduced in that PR to a separate class in the future, like the `MappingProviderUtils` seen here. (`TinyUtils` as a name is inappropriate and misleading, since MIO supports many more formats).

PR #10, opened 2019-05-26
when fixing conflicts, strip class names only when present

asiekierka:


sfPlayer1:
I merged an unconditional variation as that should be always correct, thanks

PR #13, opened 2019-09-04
Can accept proguard defintion and remap official to mojang name now

liach:
Signed-off-by: liach <liach@users.noreply.github.com>


modmuss50:
Tested the PR and it works as expected.

Would it be possible/wanted to add support to the CommandLine for it?

The rest of the PR is really upto player.

modmuss50:
I’m not actually too sure this needs to go into the main project. Being a separate lib might be better.

NebelNidas:
With #128 merged, I think this can be closed

PR #14, opened 2019-09-10
Properly map params of overriding methods

liach:
@sfPlayer1 pinging for a review.

After this change, method parameters like the tag parameter for read/write custom compound tag is now properly mapped.


liach:
@sfPayer1 updated to use immediate declaration first whenever possible.

liach:
fyi just force pushed to remove extra braces for an `if` statement

sfPlayer1:
Thanks!

PR #15, opened 2019-09-13
Fixes difference between LVT and parameters

NeunEinser:
See https://github.com/FabricMC/tiny-remapper/commit/c97e00177f7c7a5c6da159d6d2faa09d4c32875e#r35064880

Previously:
```
    Signature:
      <T:Ljava/lang/Object;>(Lnet/minecraft/util/registry/Registry<TT;>;Ljava/util/function/Function<TT;Ljava/lang/String;>;Ljava/util/Set<Ljava/lang/String;>;)V
    MethodParameters:
      Name  Flags  
      ----  -----
      null  
      null  
      null  
    LocalVariableTable:
      Start  Length  Slot  Name        Signature
      -----  ------  ----  ----------  --------------------------------------
      0      22      0     registry_1  Lnet/minecraft/util/registry/Registry;
      0      22      1     function_1  Ljava/util/function/Function;
      0      22      2     set_1       Ljava/util/Set;
      4      18      3     language_1  Lnet/minecraft/util/Language;
```
Now:
```
    Signature:
      <T:Ljava/lang/Object;>(Lnet/minecraft/util/registry/Registry<TT;>;Ljava/util/function/Function<TT;Ljava/lang/String;>;Ljava/util/Set<Ljava/lang/String;>;)V
    MethodParameters:
      Name        Flags  
      ----------  -----
      registry_1  
      function_1  
      set_1       
    LocalVariableTable:
      Start  Length  Slot  Name        Signature
      -----  ------  ----  ----------  --------------------------------------
      0      22      0     registry_1  Lnet/minecraft/util/registry/Registry;
      0      22      1     function_1  Ljava/util/function/Function;
      0      22      2     set_1       Ljava/util/Set;
      4      18      3     language_1  Lnet/minecraft/util/Language;
```

Previously:
```
  protected abstract void method_21112(int p0, int p1, boolean p2);
    Flags: PROTECTED, ABSTRACT
    MethodParameters:
      Name  Flags  
      ----  -----
      null  
      null  
      null  
```

Now:
```
  protected abstract void method_21112(int int_1, int int_2, boolean boolean_1);
    Flags: PROTECTED, ABSTRACT
    MethodParameters:
      Name       Flags  
      ---------  -----
      int_1      
      int_2      
      boolean_1  
```

### DIFF
--- a/src/main/java/net/fabricmc/tinyremapper/AsmClassRemapper.java
+++ b/src/main/java/net/fabricmc/tinyremapper/AsmClassRemapper.java
@@ -129,14 +129,12 @@ class AsmClassRemapper extends ClassRemapper {
 		public void visitParameter(String name, int access) {
 			name = ((AsmRemapper) remapper).mapMethodArg(this.owner, this.name, this.desc, getLvIndex(this.desc, isStatic, argsVisited), name);
 
-			if (renameInvalidLocals) {
-				if (!isValidJavaIdentifier(name)) {
-					name = renamedInvalidLocals.getOrDefault(argsVisited, name);
-				}
-
-				if (this.isAbstract && !isValidJavaIdentifier(name)) {
+			if (renameInvalidLocals && !isValidJavaIdentifier(name)) {
+				if (this.isAbstract) {
 					Type type = Type.getArgumentTypes(this.desc)[argsVisited];
 					name = getGeneratedName(type);
+				} else {
+					name = renamedInvalidLocals.getOrDefault(argsVisited, name);
 				}
 			}
 
@@ -161,7 +159,7 @@ class AsmClassRemapper extends ClassRemapper {
 
 			int argCount = Type.getArgumentTypes(this.desc).length;
 
-			for (int i = argsVisited; i < argCount; i++) {
+			for (int i = argsVisited; i < argCount && argsVisited == i; i++) {
 				visitParameter(null, 0);
 			}
 		}
@@ -300,9 +298,9 @@ class AsmClassRemapper extends ClassRemapper {
 			return bsm.getTag() == Opcodes.H_INVOKESTATIC
 					&& bsm.getOwner().equals("java/lang/invoke/LambdaMetafactory")
 					&& (bsm.getName().equals("metafactory")
-							&& bsm.getDesc().equals("(Ljava/lang/invoke/MethodHandles$Lookup;Ljava/lang/String;Ljava/lang/invoke/MethodType;Ljava/lang/invoke/MethodType;Ljava/lang/invoke/MethodHandle;Ljava/lang/invoke/MethodType;)Ljava/lang/invoke/CallSite;")
-							|| bsm.getName().equals("altMetafactory")
-							&& bsm.getDesc().equals("(Ljava/lang/invoke/MethodHandles$Lookup;Ljava/lang/String;Ljava/lang/invoke/MethodType;[Ljava/lang/Object;)Ljava/lang/invoke/CallSite;"))
+					&& bsm.getDesc().equals("(Ljava/lang/invoke/MethodHandles$Lookup;Ljava/lang/String;Ljava/lang/invoke/MethodType;Ljava/lang/invoke/MethodType;Ljava/lang/invoke/MethodHandle;Ljava/lang/invoke/MethodType;)Ljava/lang/invoke/CallSite;")
+					|| bsm.getName().equals("altMetafactory")
+					&& bsm.getDesc().equals("(Ljava/lang/invoke/MethodHandles$Lookup;Ljava/lang/String;Ljava/lang/invoke/MethodType;[Ljava/lang/Object;)Ljava/lang/invoke/CallSite;"))
 					&& !bsm.isInterface();
 		}
 
--- a/src/main/java/net/fabricmc/tinyremapper/AsmClassRemapper.java
+++ b/src/main/java/net/fabricmc/tinyremapper/AsmClassRemapper.java
@@ -104,6 +104,7 @@ class AsmClassRemapper extends ClassRemapper {
 			this.name = name;
 			this.desc = desc;
 			this.isStatic = (access & Opcodes.ACC_STATIC) != 0;
+			this.isAbstract = (access & Opcodes.ACC_ABSTRACT) != 0;
 			this.argLvSize = getLvIndex(desc, isStatic, Integer.MAX_VALUE);
 			this.renameInvalidLocals = renameInvalidLocals;
 		}
@@ -125,14 +126,26 @@ class AsmClassRemapper extends ClassRemapper {
 		}
 
 		@Override
-		public void visitParameter(String name, int access) {
-			name = ((AsmRemapper) remapper).mapMethodArg(this.owner, this.name, this.desc, getLvIndex(this.desc, isStatic, argsVisited), name);
-			argsVisited++;
-			super.visitParameter(name, access);
+		public void visitParameter(final String name, int access) {
+			String newName = ((AsmRemapper) remapper).mapMethodArg(this.owner, this.name, this.desc, getLvIndex(this.desc, isStatic, argsVisited), name);
+
+			if (newName == null || newName.equals(name)) {
+				newName = renamedInvalidLocals.get(argsVisited);
+			}
+
+			if(this.isAbstract && (newName == null || newName.equals(name))) {
+				newName = "p_" + nameCounts.compute("p", (k, v) -> (v == null) ? 1 : v + 1);
+			}
+
+			if(newName != null) {
+				argsVisited++;
+				super.visitParameter(newName, access);
+			}
 		}
 
-		private void checkParameters() {
-			if (argsVisited > 0 || this.desc.startsWith("()")) return;
+		private void checkParameters(boolean forceCheck) {
+			if ((checkedParams || this.desc.startsWith("()")) && !forceCheck) return;
+			checkedParams = true;
 
 			/* visitParameter() hasn't been called. checkParameters() gets called for every visit method that may
 			 * follow visitParameters(), but never earlier, in accordance with the visit order constraints specified in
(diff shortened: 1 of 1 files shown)


NeunEinser:
Pinging @sfPlayer1 for a review.

src/main/java/net/fabricmc/tinyremapper/AsmClassRemapper.java
```diff
@@ -125,14 +126,26 @@ private static int getLvIndex(String desc, boolean isStatic, int asmIndex) {
...
-			super.visitParameter(name, access);
+		public void visitParameter(final String name, int access) {
+			String newName = ((AsmRemapper) remapper).mapMethodArg(this.owner, this.name, this.desc, getLvIndex(this.desc, isStatic, argsVisited), name);
+
+			if (newName == null || newName.equals(name)) {
+				newName = renamedInvalidLocals.get(argsVisited);
+			}
+
+			if(this.isAbstract && (newName == null || newName.equals(name))) {
+				newName = "p_" + nameCounts.compute("p", (k, v) -> (v == null) ? 1 : v + 1);
```
Chocohead: The type of a parameter can be inferred from the method description as per [the spec](https://docs.oracle.com/javase/specs/jvms/se8/html/jvms-4.html#jvms-4.7.24):
> The i'th entry in the parameters array corresponds to the i'th parameter descriptor in the enclosing method's descriptor.

Taking the assumption that `argsVisited` is never erroneously bumped more (or indeed less) than necessary, that provides `i` in `Type.getArgumentTypes(this.desc)` which could then be handled as `#visitLocalVariable` does.

Restricting this to abstract methods only is slightly questionable too as an LVT doesn't have to be present on a normal method (as Mojang indeed didn't until 1.8.4). There won't be a direct way to know whether a method does or doesn't have one until `#visitEnd` is reached as at that point all the method information that the class has will have been passed through. Perhaps it is best that `#checkParameters` is only called once from `#visitEnd` so that any parameters names from the LVT have definitely been passed?

src/main/java/net/fabricmc/tinyremapper/AsmClassRemapper.java
```diff
@@ -125,14 +126,26 @@ private static int getLvIndex(String desc, boolean isStatic, int asmIndex) {
...
+			}
+
+			if(newName != null) {
```
Chocohead: For names carried over directly from the class file which aren't immediately remapped or renamed by the LVT (presuming a non-abstract method), it would be nice to keep the access modifier they might have even if the name is thrown away

src/main/java/net/fabricmc/tinyremapper/AsmClassRemapper.java
NeunEinser: > The type of a parameter can be inferred from the method description
Already did that in a newer commit. Thanks for pointing it out anyways :)

> Restricting this to abstract methods only is slightly questionable too as an LVT doesn't have to be present on a normal method (as Mojang indeed didn't until 1.8.4). There won't be a direct way to know whether a method does or doesn't have one until `#visitEnd` is reached as at that point all the method information that the class has will have been passed through. Perhaps it is best that `#checkParameters` is only called once from `#visitEnd` so that any parameters names from the LVT have definitely been passed?
Sure, I could do that.

src/main/java/net/fabricmc/tinyremapper/AsmClassRemapper.java
NeunEinser: In the newest commit, it keeps it always, when renameinavlidmodifier is not set. Otherwise, it only keeps it when we found a valid identifier.
I am not sure how to keep the access modifier otherwise, perhabs just push it to a list every time the method is called and later retrieve it from there?

Chocohead: Until `checkedParams` is `true` you can be sure that the visits are coming from the `ClassReader` rather than internally, at which point if `access != 0` it could be stored in an `int[]` with length `argCount`.

Then once `checkedParams` is `true`, `#checkParameters` could pull the access from the array rather than always using `0`. Although this might need a separate way of watching the end of the array up to this point if `argsVisited` isn't always bumped.

src/main/java/net/fabricmc/tinyremapper/AsmClassRemapper.java
```diff
@@ -198,20 +212,13 @@ public void visitLocalVariable(String name, String descriptor, String signature,
 
 			if (renameInvalidLocals && !isValidJavaIdentifier(name)) {
 				Type type = Type.getType(descriptor);
-				boolean plural = false;
+				name = getGeneratedName(type);
```
Chocohead: It would be nice if this could be pulled from the appropriate parameter slot if it had already been visited (as that would also mean it was valid), although that would need all the `super.visitParameter` calls to have been saved to an array too to know.

NeunEinser: Oh, so you are saying, if visitParameter didn't find the name, just generate one there and then retrieve it here.
Yeah, that makes a lot of sense and actually also makes the access modifier thing a bit cleaner, too, since we can just always map at the end of visitParameter then.

src/main/java/net/fabricmc/tinyremapper/AsmClassRemapper.java
NeunEinser: I'm now using a different approach where the parameter is guaranteed when the option is set, so this is no longer a problem.

src/main/java/net/fabricmc/tinyremapper/AsmClassRemapper.java
```diff
@@ -127,26 +126,20 @@ private static int getLvIndex(String desc, boolean isStatic, int asmIndex) {
...
+				final Type type = Type.getArgumentTypes(this.desc)[argsVisited];
+				name = getNameFromType(type, lvIndex);
 			}
 
-			if(!renameInvalidLocals || isValidJavaIdentifier(name)) {
-				argsVisited++;
-				super.visitParameter(name, access);
-			}
+			argsVisited++;
+			super.visitParameter(name, access);
```
Chocohead: This will allow `null` parameter names if `name` is null (which as we've seen before end up as `paramx` or `<anonymous>`). Might want to null check `name` but still increase `argsVisited` regardless given `renameInvalidLocals` will be false to get into the situation.

src/main/java/net/fabricmc/tinyremapper/AsmClassRemapper.java
```diff
@@ -244,7 +235,10 @@ private String getGeneratedName(Type type) {
 
 			varName = Character.toLowerCase(varName.charAt(0)) + varName.substring(1);
 			if (plural) varName += "s";
-			return varName + "_" + nameCounts.compute(varName, (k, v) -> (v == null) ? 1 : v + 1);
+			String name = varName + "_" + nameCounts.compute(varName, (k, v) -> (v == null) ? 1 : v + 1);
+			renamedInvalidLocals.put(index, name);
```
Chocohead: This isn't safe for non-parameter LVT indexes, they can and will be reused:
```java
public void method(int param) {
	if (...) {
		int value = ...
		...
	}

	if (...) {
		String otherValue = ...
		...
	}
}
```
would end up being remapped to
```java
public void method(int int_1) {
	if (...) {
		int int_2 = ...
		...
	}

	if (...) {
		String int_2 = ...
		...
	}
}
```
Perhaps they should only be added when `index < argCount`? You could also do `String[] renamedInvalidLocals  = new String[argCount]` then check if `index < renamedInvalidLocals.length`.

src/main/java/net/fabricmc/tinyremapper/AsmClassRemapper.java
NeunEinser: Yes, that is by design, since we might still have mappings for parameters down the line. I don't think it really hurts to produce something invalid when the option is not set, as there won't be any change in the LVT as well.

Not doing the call when there's a null value is definitely wrong, since the parameters get added in the order they are encountered.
This would mean, that if we have no mapping for the first parameter, the second parameter's mapping might be applied for the first parameter.

Chocohead: I was thinking that the LVT might have a name that could be used instead, but I suppose if you've got `renameInvalidLocals` false consistency between the two is probably not a problem given the lack of concern about the LVT being valid to start with.

NeunEinser: One thing I could do in `checkParameters` is to first iterate all mappings and see if there is at least one, and only if I find a parameter mapping, actually call `visitParameter`.
That way no mapping happens when we have absolutely none.

src/main/java/net/fabricmc/tinyremapper/AsmClassRemapper.java
NeunEinser: >  You could also do `String[] renamedInvalidLocals = new String[argCount]` then check if `index < renamedInvalidLocals.length`.

I could do that, but I don't know if that's actually more efficient. This would mean, that I have to calculate the index in `visitLocalVariable`. Currently, the lvIndex needs to be calculated in `checkParameters`, but that gets calculated either way, since it's needed for the parameter name mapping method.

Chocohead: Oh yeah double slot parameters would mean `argCount <= argLvSize`. It's probably easiest to just use `argLvSize` as the array size with holes (given there'll never be any locals which would use the holes). Saves having to do any more work for LVT indexes than is done already.

src/main/java/net/fabricmc/tinyremapper/AsmClassRemapper.java
Chocohead: That's a good middle ground solution for the situation, it shouldn't slow down remapping too much given the maximum number of parameters is only 255, and it's only when `renameInvalidLocals` is false.

NeunEinser: As Player pointed out on Discord, we need to also look for a valid LVT parameter first.
So I will need to hold the entire parameter generation back to after all LVT names are processed.

Because of this, I'll probably move the mapping to visitLocalVariable entirely and at the end just take whatever this method used.

This will also require to save the access modifier again.

NeunEinser: In what I just pushed, we get this without renaming invalids:
```
    MethodParameters:
      Name  Flags  
      ----  -----
      \u2603     
      \u2603     
      \u2603     
      \u2603     
    LocalVariableTable:
      Start  Length  Slot  Name  Signature
      -----  ------  ----  ----  --------------------------
      0      26      0     this  Lnet/minecraft/class_4371;
      0      26      1     \u2603     I
      0      26      2     \u2603     I
      0      26      3     \u2603     I
      0      26      4     \u2603     I
```

And in an abstract method without renaming invalids:
```
  protected abstract void method_21112(int p0, int p1, boolean p2);
    Flags: PROTECTED, ABSTRACT
    MethodParameters:
      Name  Flags  
      ----  -----
      null  
      null  
      null  
```

Of course parameters that actually do have a mapping are not affected:
```
    Signature:
      <T:Ljava/lang/Object;>(Lnet/minecraft/util/registry/Registry<TT;>;Ljava/util/function/Function<TT;Ljava/lang/String;>;Ljava/util/Set<Ljava/lang/String;>;)V
    MethodParameters:
      Name             Flags  
      ---------------  -----
      registry         
      keyExtractor     
      translationKeys  
    LocalVariableTable:
      Start  Length  Slot  Name             Signature
      -----  ------  ----  ---------------  --------------------------------------
      0      22      0     registry         Lnet/minecraft/util/registry/Registry;
      0      22      1     keyExtractor     Ljava/util/function/Function;
      0      22      2     translationKeys  Ljava/util/Set;
      4      18      3     \u2603                Lnet/minecraft/util/Language;
```

PR #24, opened 2019-12-23
Fix member descriptors when the from namespace isn't first

Chocohead:
Previously mapped member descriptors to the `to` namespace instead, which isn't how `IMappingProvider` is designed to take them (as it would be odd to have the `from` namespace name with the `to` namespace descriptor).


sfPlayer1:
Thanks!

PR #30, opened 2020-01-12
Prevent keywords or literals being used as local var names

Marcono1234:
Previously `AsmClassRemapper` could create local var names which are keywords or literals, or it would not consider existing variable names which are keywords or literals as invalid.

It now uses [SourceVersion.isName](https://docs.oracle.com/javase/8/docs/api/javax/lang/model/SourceVersion.html#isName-java.lang.CharSequence-) to detect invalid names, whose behavior sadly depends on the running Java version. In Java 9(+) this can be solved by using the [method overload](https://docs.oracle.com/en/java/javase/11/docs/api/java.compiler/javax/lang/model/SourceVersion.html#isName(java.lang.CharSequence,javax.lang.model.SourceVersion)) which accepts a `SourceVersion`.

[test.zip](https://github.com/FabricMC/tiny-remapper/files/4051303/test.zip) contains a class file for which tiny-remapper would previously create invalid parameter names.

### DIFF
--- a/src/main/java/net/fabricmc/tinyremapper/AsmClassRemapper.java
+++ b/src/main/java/net/fabricmc/tinyremapper/AsmClassRemapper.java
@@ -456,13 +456,23 @@ class AsmClassRemapper extends ClassRemapper {
 				throw new IllegalStateException();
 			}
 
-			if (plural) varName += "s";
+			boolean hasPluralS = false;
+
+			if (plural) {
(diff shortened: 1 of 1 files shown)


Marcono1234:
Good point, thanks! I have added a TODO comment for replacing `SourceVersion.isKeyword` with the one accepting a `SourceVersion` in Java 9. I hope that is alright, but I can also remove it again if you want.

PR #31, opened 2020-04-02
ASM 8.0

modmuss50:


liach:
mind find if there is asm4/asm5 etc usages and replace them as well

modmuss50:
I did

PR #33, opened 2020-05-01
Append '_' to local var names in case they are keywords

Marcono1234:
Previously local variables and arguments whose generated name was a keyword would instead get the unspecific names "lv" or "arg" (see also #30).
This pull request changes it so the generated name followed by a '_' (similar to how CFR does it) is used.

With `_` becoming a keyword in Java 9, `_` would become `__` (which is valid). Should this get a special case?

### DIFF
--- a/src/main/java/net/fabricmc/tinyremapper/AsmClassRemapper.java
+++ b/src/main/java/net/fabricmc/tinyremapper/AsmClassRemapper.java
@@ -252,7 +252,8 @@ class AsmClassRemapper extends ClassRemapper {
 						int asmIndex = getAsmIndex(lv.index, isStatic, argTypes);
 						String existingName = args[asmIndex];
 
-						if (existingName == null || !isValidJavaIdentifier(existingName) && isValidJavaIdentifier(lv.name)) { // replace if missing or better
+						// replace if missing or better; don't check for keywords, will be renamed later
+						if (existingName == null || !isValidJavaIdentifier(existingName) && isValidJavaIdentifier(lv.name)) {
 							args[asmIndex] = lv.name;
 						}
 
@@ -268,7 +269,7 @@ class AsmClassRemapper extends ClassRemapper {
 
 							lv.name = ((AsmRemapper) remapper).mapMethodVar(owner, methodNode.name, methodNode.desc, lv.index, startOpIdx, i, lv.name);
 
-							if (renameInvalidLocals && isValidJavaIdentifier(lv.name)) { // block valid name from generation
+							if (renameInvalidLocals && isValidJavaIdentifierAndNotKeyword(lv.name)) { // block valid name from generation
 								nameCounts.putIfAbsent(lv.name, 1);
 							}
 						}
@@ -283,7 +284,7 @@ class AsmClassRemapper extends ClassRemapper {
 				for (int i = 0; i < args.length; i++) {
 					args[i] = ((AsmRemapper) remapper).mapMethodArg(owner, methodNode.name, methodNode.desc, getLvIndex(i, isStatic, argTypes), args[i]);
 
-					if (renameInvalidLocals && isValidJavaIdentifier(args[i])) { // block valid name from generation
+					if (renameInvalidLocals && isValidJavaIdentifierAndNotKeyword(args[i])) { // block valid name from generation
 						nameCounts.putIfAbsent(args[i], 1);
 					}
 				}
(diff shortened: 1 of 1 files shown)


src/main/java/net/fabricmc/tinyremapper/AsmClassRemapper.java
```diff
@@ -481,7 +482,21 @@ private String getNameFromType(String type, boolean isArg) {
...
+					varName = varName.concat(Integer.toString(count));
+				}
+				/*
+				 * Perform keyword check here, because appending numbers for duplicate names
+				 * should prevent keywords, so result will be:
+				 *   class_, class2, class3
+				 * If keyword check was performed before checking for duplicates, result would be:
+				 *   classs_, class_2, class_3
+				 */
+				else if (getInvalidJavaIdentifierReason(varName) == InvalidIdentifierReason.KEYWORD) {
```
sfPlayer1, line 495: I'd use something like `} else {\n/* long\n comment*/ if (getInvalidJavaIdent[...]) {\n[...]\n}\n}` to retain the formatting style.

src/main/java/net/fabricmc/tinyremapper/AsmClassRemapper.java
```diff
@@ -516,11 +531,28 @@ private static String getIndexName(int index, boolean plural) {
...
+			KEYWORD
+		}
+
+		/**
+		 * @param s
+		 *        Name of the identifier
+		 * @return
+		 *        Reason why the identifier is invalid, or {@code null} if it is valid
+		 */
+		private static InvalidIdentifierReason getInvalidJavaIdentifierReason(String s) {
```
sfPlayer1, line 545: This method is likely better off being split into something like isValidOrKeyword and isKeyword without the enum. Most uses should be fine with checking invalid once, then only isKeyword later (e.g. lines 466, 475).

src/main/java/net/fabricmc/tinyremapper/AsmClassRemapper.java
```diff
@@ -481,7 +482,21 @@ private String getNameFromType(String type, boolean isArg) {
...
+				}
+				/*
+				 * Perform keyword check here, because appending numbers for duplicate names
+				 * should prevent keywords, so result will be:
+				 *   class_, class2, class3
+				 * If keyword check was performed before checking for duplicates, result would be:
+				 *   classs_, class_2, class_3
+				 */
+				else if (getInvalidJavaIdentifierReason(varName) == InvalidIdentifierReason.KEYWORD) {
+					varName += '_';
```
sfPlayer1, line 496: This new name can clash with existing variables, like when user-provided through the mapping file. It needs to do the same nameCount operation as a few lines above and potentially fall back to not append _ and use the appropriate integer suffix instead.

On further look there appears to be a similar collision problem with pre-existing integer suffixed names (not generated by TR, but from mappings or original LVT) in the old code. The various `nameCounts.putIfAbsent` invocations that pre-populate the index with pre-existing names don't split those names into base-name and suffix. This way nameCounts may have an entry `(class2, 1)` instead of `(class, 2)` after recording `class2`. This then causes the name generator to turn `class` into `class_`, `class2`, `class3` instead of `class3`, `class4`, `class5`. I don't know if you want to fix this or if I should instead.

src/main/java/net/fabricmc/tinyremapper/AsmClassRemapper.java
Marcono1234: Thanks, that comment does not exist in the new version anymore

src/main/java/net/fabricmc/tinyremapper/AsmClassRemapper.java
Marcono1234: Tried to apply that in the new version

src/main/java/net/fabricmc/tinyremapper/AsmClassRemapper.java
Marcono1234: Good point, should be fixed in the new version.

Not recording the name with the count suffix should be fixed now. However, I did not implement splitting of mapped names ending with a number. Not sure if it would really matter which variables got which number, and trying to determine the meaning of a number suffix could also behave incorrectly, e.g. when the class name ends with a number: `MyClass2`

Marcono1234:
Thanks for this feedback! I hopefully implemented all requested changes now and forced pushed them. Could you please have a look?

PR #34, opened 2020-05-01
Use short variable names for boxed primitives

Marcono1234:
Currently there is no special handling for boxed primitives (such as `java.lang.Integer`) so variables of these types get the class name, e.g. `integer`.
This pull request changes it so they get the same name as the respective primitive type, e.g. `java.lang.Integer` -> `i`.

However, maybe that is not desired because a variable named `integer` indiciates that it might be `null`, while that is not the case for `i`. What do you think?

### DIFF
--- a/src/main/java/net/fabricmc/tinyremapper/AsmClassRemapper.java
+++ b/src/main/java/net/fabricmc/tinyremapper/AsmClassRemapper.java
@@ -410,19 +410,23 @@ class AsmClassRemapper extends ClassRemapper {
 			boolean incrementLetter = true;
 			String varName;
 
-			switch (type.charAt(0)) {
-			case 'B': varName = "b"; break;
-			case 'C': varName = "c"; break;
-			case 'D': varName = "d"; break;
-			case 'F': varName = "f"; break;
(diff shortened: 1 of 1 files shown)


sfPlayer1:
IMO the short names shine in frequent use areas. The primitive types are everywhere for indexing, coordinate/position/offset tuples, in expressions, short lived variables etc. where long names differing only in the last character hurt comprehension. The boxed forms don't usually share these applications and are hardly seen at all, while as you noted the nullability can be quite important to highlight.

With that in mind I believe most code wouldn't benefit, counter-examples are welcome though. This translation scheme might be another candidate for an opt-in setting.

liach:
Imo we don't need the special handling. Most primitive usages in java code is unboxed, and boxed version is usually immediately boxed/unboxed when converting between primitive and object.

Marcono1234:
There are cases where using the short name might make sense, e.g.:
- lambdas
- collections
- In Minecraft: Usage of `com.mojang.datafixers.util.Pair`

However, there you have no indication whether a value might be null, so the long name would be appropriate.

I will therefore close this pull request. Thanks for your feedback!

PR #35, opened 2020-05-02
Fix variable names containing ignorable chars not being renamed

Marcono1234:
Ignorable chars (`Character.isIdentifierIgnorable`) are ignored as part of identifiers (cannot be leading, but can be in between and trailing) when compiling. They can therefore normally not appear in a class file unless it was modified. Since they cannot be represented in source (and could easily cause identifier collisions), identifiers containing them are now considered invalid.

This was only added to the JLS for Java 10 (see [JDK-6214519](https://bugs.openjdk.java.net/browse/JDK-6214519)), however the behavior also exists in earlier versions.

Demonstration of the issue:
1. Download [IdentifierIgnorable.class.txt](https://github.com/FabricMC/tiny-remapper/files/4566792/IdentifierIgnorable.class.txt) (remove `.txt`, that is only to make GitHub happy)
2. Use Tiny Remapper and let it rename invalid variable names

### DIFF
--- a/src/main/java/net/fabricmc/tinyremapper/AsmClassRemapper.java
+++ b/src/main/java/net/fabricmc/tinyremapper/AsmClassRemapper.java
@@ -545,7 +545,10 @@ class AsmClassRemapper extends ClassRemapper {
 		}
 
 		private static boolean isValidJavaIdentifier(String s) {
-			return s != null && !s.isEmpty() && SourceVersion.isIdentifier(s);
+			return s != null && !s.isEmpty() && SourceVersion.isIdentifier(s)
+					// Ignorable characters cannot be represented in source code,
+					// would be ignored when re-compiled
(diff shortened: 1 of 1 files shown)


Marcono1234:
Force pushed commit which solves the conflict.
`String.codePoints()` was indeed added in Java 9, however it only improves the performance of `CharSequence.codePoints()` (see [JDK-8071477](https://bugs.openjdk.java.net/browse/JDK-8071477)) which already exists in Java 8. The javadoc is sadly a little bit irritating with its `@since 9` there (was added due to [JDK-8189952](https://bugs.openjdk.java.net/browse/JDK-8189952)).
So unless that perfomance difference is an issue, I would like to use `codePoints()`.

sfPlayer1:
Right, that was sneaky. It may be a rather busy method, so not sure if the extra time or gc pressure matters.

PR #38, opened 2020-05-25
Implement hashCode and equals methods on Member

Nuclearfarts:
Implement the `hashCode` and `equals` methods on `Member`. This makes using `Member` as a key in a HashMap possible.

### DIFF
--- a/src/main/java/net/fabricmc/tinyremapper/IMappingProvider.java
+++ b/src/main/java/net/fabricmc/tinyremapper/IMappingProvider.java
@@ -39,5 +39,20 @@ public interface IMappingProvider {
 		public String owner;
 		public String name;
 		public String desc;
+
+		@Override
+		public int hashCode() {
+			return (31 * (31 * owner.hashCode()) + name.hashCode()) + desc.hashCode();
+		}
+
+		@Override
+		public boolean equals(Object other) {
+			if (!(other instanceof Member)) {
+				return false;
+			} else {
+				Member otherMember = (Member) other;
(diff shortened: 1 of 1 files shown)


liach:
Shall they be used in hash maps, their fields should be final. Implementing these two is incomplete for hashmap usage without making the fields final.

Nuclearfarts:
Ah, hmm, I completely missed that the fields were mutable.

src/main/java/net/fabricmc/tinyremapper/IMappingProvider.java
```diff
@@ -39,5 +39,20 @@ public Member(String owner, String name, String desc) {
 		public String owner;
 		public String name;
 		public String desc;
+
+		@Override
+		public int hashCode() {
+			return (31 * (31 * owner.hashCode()) + name.hashCode()) + desc.hashCode();
```
Marcono1234, line 45: You could use [`java.util.Objects.hash`](https://docs.oracle.com/javase/8/docs/api/java/util/Objects.html#hash-java.lang.Object...-) for that

src/main/java/net/fabricmc/tinyremapper/IMappingProvider.java
```diff
@@ -39,5 +39,20 @@ public Member(String owner, String name, String desc) {
...
+		}
+
+		@Override
+		public boolean equals(Object other) {
```
Marcono1234, line 49: Maybe it should start with `if (other == this) { return true; }` for performance for that case, though I don't know if that is needed

PR #40, opened 2020-09-17
Added module support via Manifest attribute

gudenau:


build.gradle
```diff
@@ -40,7 +40,8 @@ jar {
 	manifest {
 		attributes	'Implementation-Title': 'TinyRemapper',
 					'Implementation-Version': version,
-					'Main-Class': "net.fabricmc.tinyremapper.Main"
+					'Main-Class': "net.fabricmc.tinyremapper.Main",
+					'Automatic-Module-Name': 'net.fabricmc.TinyRemapper'
```
zml2008, line 44: should probably be lower-case to match package and artifact id naming conventions

liach:
```suggestion
					'Automatic-Module-Name': 'net.fabricmc.tinyremapper'
```

gudenau:
Woops, I left this to rot and now I don't have that fork.

PR #43, opened 2020-12-28
Allow threads to be configurable from the CLI

DemonWav:
This relates to, but is technically separate from #44. That issue was the motivation for this PR.

Other than determining threading issues and being able to mitigate them when they happen, this is nice to have in general from a user experience standpoint.

I also added `.idea/` to `.gitignore` because that was annoying me.

### DIFF
--- a/src/main/java/net/fabricmc/tinyremapper/Main.java
+++ b/src/main/java/net/fabricmc/tinyremapper/Main.java
@@ -48,6 +48,7 @@ public class Main {
 		boolean skipLocalVariableMapping = false;
 		boolean renameInvalidLocals = false;
 		NonClassCopyMode ncCopyMode = NonClassCopyMode.FIX_META_INF;
+		int threads = -1;
 
 		for (String arg : rawArgs) {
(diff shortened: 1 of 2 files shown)


modmuss50 (approved):

src/main/java/net/fabricmc/tinyremapper/Main.java
```diff
@@ -104,6 +105,13 @@ public static void main(String[] rawArgs) {
 						System.exit(1);
 					}
 
+					break;
+				case "threads":
+					threads = Integer.parseInt(arg.substring(valueSepPos + 1));
+					if (threads <= 0) {
+						System.out.println("Thread count must be >= 0");
```
Juuxel, line 112: Shouldn't this be `> 0`?

DemonWav: Yeah good catch, fixed.

sfPlayer1 (approved):
LGTM, thanks!

PR #50, opened 2021-04-25
Updated gradle wrapper version to support Java 16.

gudenau:


src/main/java/net/fabricmc/tinyremapper/TinyRemapper.java
```diff
@@ -397,18 +384,46 @@ public FileVisitResult visitFile(Path file, BasicFileAttributes attrs) throws IO
 		List<ClassInstance> ret = new ArrayList<ClassInstance>();
 
 		if (file.toString().endsWith(".class")) {
-			ClassInstance res = analyze(isInput, tags, srcPath, Files.readAllBytes(file));
+			ClassInstance res = analyze(isInput, tags, srcPath, OptionalInt.empty(), Files.readAllBytes(file)); // FIXME?
```
liach, line 387: No need for fixme; mrj mechanism doesn't exist for flat directory

src/main/java/net/fabricmc/tinyremapper/TinyRemapper.java
```diff
@@ -397,18 +384,46 @@ public FileVisitResult visitFile(Path file, BasicFileAttributes attrs) throws IO
...
 			if (res != null) ret.add(res);
 		} else {
 			URI uri = new URI("jar:"+file.toUri().toString());
 			FileSystem fs = FileSystemHandler.open(uri);
 			fsToClose.add(fs);
-
+			
+			boolean isMrj = false;
+			if(file.getFileName().toString().endsWith(".jar")){
+				System.out.println("jar");
```
liach, line 396: I recall java util's jar file api has support for getting version-aware jar entries. If there is existing api, we should use those than build our own since our own is vulnerable to ignoring impl/api requirements

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,5 +1,5 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-6.7.1-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-7.0-bin.zip
```
modmuss50, line 3: You can also update the github workflows to use java 16 now.

Issue #45, opened 2021-01-26
Multi-release JARs break when remapping


ramidzkh:
1. Inspect App.jar. Notice that there are files with the same name as `App` within the `META-INF/versions` folder
2. Try running `App.jar App2.jar mappings.tiny a b`
3. The remapper complains several times
```
duplicate input class App, from App.jar and App.jar
duplicate input class App, from App.jar and App.jar
duplicate input class App, from App.jar and App.jar
duplicate input class App, from App.jar and App.jar
duplicate input class App, from App.jar and App.jar
duplicate input class App, from App.jar and App.jar
duplicate input class App, from App.jar and App.jar
duplicate input class App, from App.jar and App.jar
Finished after 398.13 ms.
```
4. Inspect the remapped jar. Notice that the classes inside `META-INF/versions` is now missing. The multi-release jar is now broken

(These files are free from copyright) [Files.zip](https://github.com/FabricMC/tiny-remapper/files/5871100/Files.zip)

sfPlayer1:
This is a problem with TR not supporting multi-release jars at this moment. It isn't super trivial to implement since it involves semi-independent classes with the same name that may use a completely different sub graph in the class hierarchy.

gudenau:
Couldn't you do this in passes?

Get the highest version view of the jar and have everything lower be read-only, do the next highest one, etc..

Seems like the simplest way to handle this.

gudenau:
I think I have come up with a version of Tiny Remapper that works MRJs, I will throw a few tests cases at it tomorrow and if it passes those I'll make a PR for it.

LogicFan:
Hi, is there anyone working on this? If not, I can have a try.

sfPlayer1:
It is quite high on the TODO list, not working on it yet though. Not sure about the best approach, might be easiest to create a completely separate model for each version, otherwise everything needs to become aware of multiple same-id class instances with their own parent/child relationships.

LogicFan:
This is finished in the `remapbridges` branch.

PR #53, opened 2021-05-23
Multi Release Jar Support

LogicFan:
This is still working in progress. Currently, the dependency resolution still does not support MRJ (i.e. they all use Java 8 version, but really they should use their own version). But anything else should work just fine.

This is related to #45


src/main/java/net/fabricmc/tinyremapper/VersionedName.java
```diff
@@ -0,0 +1,62 @@
...
+
+    public static String getMrjClsName(String clsName, OptionalInt mrjVersion) {
+        if (mrjVersion.isPresent()) {
+            return MRJ_PREFIX + File.separator + mrjVersion.getAsInt() + File.separator + clsName;
+        } else {
+            return clsName;
+        }
+    }
+
+    public static String getMrjClsName(VersionedName mrjName) {
```
liach, line 46: Why don't you make this an instance method?

src/main/java/net/fabricmc/tinyremapper/TinyRemapper.java
```diff
@@ -408,7 +401,11 @@ public FileVisitResult visitFile(Path file, BasicFileAttributes attrs) throws IO
 				@Override
 				public FileVisitResult visitFile(Path file, BasicFileAttributes attrs) throws IOException {
 					if (file.toString().endsWith(".class")) {
-						ClassInstance res = analyze(isInput, tags, srcPath, Files.readAllBytes(file));
+						OptionalInt mrjVersion = OptionalInt.empty();
```
liach, line 404: Instead of using file start with check, you should pull it out so you walk the META-INF and other directories separately

src/main/java/net/fabricmc/tinyremapper/TinyRemapper.java
```diff
@@ -714,11 +711,11 @@ private void handleConflicts() {
 		}
 	}
 
-	public void apply(final BiConsumer<String, byte[]> outputConsumer) {
+	public void apply(final BiConsumer<VersionedName, byte[]> outputConsumer) {
 		apply(outputConsumer, (InputTag[]) null);
 	}
 
-	public void apply(final BiConsumer<String, byte[]> outputConsumer, InputTag... inputTags) {
+	public void apply(final BiConsumer<VersionedName, byte[]> outputConsumer, InputTag... inputTags) {
```
liach, line 718: Need to keep `BiConsumer<String, byte[]>` versions or this is a breaking change to all users

liach (changes requested):

src/main/java/net/fabricmc/tinyremapper/TinyRemapper.java
LogicFan: I do not see the advantage of walk the META-INF separately. And the current way is much easier to implement.

src/main/java/net/fabricmc/tinyremapper/TinyRemapper.java
LogicFan: fixed

src/main/java/net/fabricmc/tinyremapper/TinyRemapper.java
liach: This current way is bad performance-wise as it tries to parse the version multiple times while a walk within the META-INF can find all versions at once and avoid creating OptionalInt instances repeatedly.

LogicFan: Yes, but that will increase the complexity of code (and I don't think the current code will create too many additional OptionalInt anyway). I will leave it for now. My current priority is to fix the class dependency resolution (i.e. using the correct version instead of using the Java 8 version for everything).

liach: Well, that doesn't. That actually makes things clearer as you already have to build a class graph for each release present, and splitting the version-specific and common release file visit code will increase clarity and make code more maintainable in the future, not to mention the performance boost it is already bringing.

src/main/java/net/fabricmc/tinyremapper/VersionedName.java
liach: You should declare it like `public String getClassName()` so you can just call `.getClassName()` on your versioned name objects. Also please use full names like `MultiReleaseClass` than `MrjCls`

LogicFan: fixed

liach:
Now that I see, I recommend using a negative integer to indicate absent version

src/main/java/net/fabricmc/tinyremapper/util/VersionAwareMap.java
```diff
@@ -0,0 +1,373 @@
...
+     * @return a set view of the mappings contained in this map
+     */
+    @Override
+    public Set<Entry<VersionedName, ClassInstance>> entrySet() {
+        return values().stream()
+                .map(cls -> new AbstractMap.SimpleEntry<>(cls.getVersionedName(), cls))
+                .collect(Collectors.toSet());
+    }
+
+    public Set<Integer> versions() {
```
liach, line 364: this implementation would be inefficient if this method is called frequently. just be careful.

src/main/java/net/fabricmc/tinyremapper/util/VersionAwareMap.java
```diff
@@ -0,0 +1,373 @@
...
+     *                                       prevents it from being stored in this map
+     * @throws NullPointerException          if the specified key or value is null
+     *                                       and this map does not permit null keys or values
+     * @throws IllegalArgumentException      if some property of the specified key
+     *                                       or value prevents it from being stored in this map
+     */
+    @Override
+    public ClassInstance put(VersionedName key, ClassInstance value) {
+        ClassInstance result = remove(key);
+        storage.get(key.getName()).add(value);
```
liach, line 217: Need to prevent having duplicate versions in one entry array list.

src/main/java/net/fabricmc/tinyremapper/util/VersionAwareMap.java
```diff
@@ -0,0 +1,373 @@
...
+     * @return the previous value associated with {@code key}, or
+     * {@code null} if there was no mapping for {@code key}.
+     * @throws UnsupportedOperationException if the {@code remove} operation
+     *                                       is not supported by this map
+     * @throws ClassCastException            if the key is of an inappropriate type for
+     *                                       this map
+     *                                       (<a href="{@docRoot}/java.base/java/util/Collection.html#optional-restrictions">optional</a>)
+     * @throws NullPointerException          if the specified key is null and this
+     *                                       map does not permit null keys
+     *                                       (<a href="{@docRoot}/java.base/java/util/Collection.html#optional-restrictions">optional</a>)
```
liach, line 249: Use `{@inheritDoc}` than copying all these javadocs

src/main/java/net/fabricmc/tinyremapper/util/VersionAwareMap.java
```diff
@@ -0,0 +1,373 @@
...
+     */
+    @Override
+    public ClassInstance remove(Object key) {
+        ClassInstance result = get(key);
+
+        if (key instanceof VersionedName) {
+            String name = ((VersionedName) key).getName();
+            storage.get(name).removeIf(cls -> cls.getVersionedName().equals(key));
```
liach, line 257: Instead of calling `get`, you should just do:
```java
ClassInstance[] removed = new ClassInstance[1];
storage.get(name).removeIf(cls -> {
  if (cls.getVersionedName().equals(key)) {
    removed[0] = cls;
    return true;
  }
  return false;
});
return removed[0];
```
this is better in terms of performance.

src/main/java/net/fabricmc/tinyremapper/util/VersionAwareMap.java
LogicFan: If a duplicate version exists, then it will be removed before put.

src/main/java/net/fabricmc/tinyremapper/util/VersionAwareMap.java
LogicFan: fixed

src/main/java/net/fabricmc/tinyremapper/util/VersionAwareMap.java
LogicFan: This should only be called once pre `refresh` call, so I think it should be fine.

src/main/java/net/fabricmc/tinyremapper/util/VersionAwareMap.java
LogicFan: I agree. I will do this performance patch after everything is working.

src/main/java/net/fabricmc/tinyremapper/util/VersionAwareMap.java
liach: I still think you should make private methods that removes locally given the array list. That way, you don't need to look into the storage map twice, for performance and cleanness concerns.

src/main/java/net/fabricmc/tinyremapper/util/VersionAwareMap.java
liach: Or you can use a for loop to find and remove the object. for loop would actually be faster because we are sure there is only one matching element and we can break the loop early once we found it

src/main/java/net/fabricmc/tinyremapper/util/VersionAwareMap.java
LogicFan: Anything related to performance will be not be considered be this thing works. But I will definitely consider your suggestion when I do the final code clean-up.

sfPlayer1:
Please avoid extra runtime libraries, wildcard imports and inefficient features like streams.

I'd do the bulk of the mrj map handling by using two maps: The existing one for regular classes and another map for classes from meta-inf's versioned directory. The latter can be `Map<String, Map<Integer, ClassInstance>>` or `Map<VersionedName, ClassInstance>`. A flag on ClassInstance may be useful to quickly determine the potential existance of a mrj sibling (true for any version).

The loadMappings step is independent from your changes, so can be done once initially.

Also check the remapbridges branch out, it is the real development frontier, pending merge into master with a fairly large amount of changes.

LogicFan:
> Please avoid extra runtime libraries, wildcard imports and inefficient features like streams.
> 
> I'd do the bulk of the mrj map handling by using two maps: The existing one for regular classes and another map for classes from meta-inf's versioned directory. The latter can be `Map<String, Map<Integer, ClassInstance>>` or `Map<VersionedName, ClassInstance>`. A flag on ClassInstance may be useful to quickly determine the potential existance of a mrj sibling (true for any version).
> 
> The loadMappings step is independent from your changes, so can be done once initially.
> 
> Also check the remapbridges branch out, it is the real development frontier, pending merge into master with a fairly large amount of changes.

I agree with most of what you said. However, there are two things I want to get clarified.

1. The usage of `Multimap` would dramatically decrease the complexity of `VersionAwareMap` and I doubt if we should strictly forbid the usage of extra libraries. Do we really want to reinvent the wheel?
2. The current way of handling the MRJ is to separate MRJ into multiple normal JAR and process them independently, then finally merge them together. I think this way would impact the least part of the current implementation. I do not quite understand how two maps & MRJ flags would work out.

liach:
Multimap is actually bad as it will create empty array lists in the map for a query to a key with empty value, even though it would not be necessary, I recall.

dutingda:
Hi, I have this issue that related to MRJs remapping also, this seems to be quite a big problem for some tasks that I have been doing right now.

sfPlayer1:
Actually I think a 3 map approach is the easiest and still fairly quick:
1. `Map<String, ClassInstance>` for the non-META-INF classes only (existing `readClasses`)
2. `Map<Integer, Map<String, ClassInstance>>` for the versioned META-INF classes only (new `readMrjClasses`)
3. `Map<Integer, MrjState>` for all mrj versions (base+whatever meta-inf discriminates between), `MrjState` replaces/holds `TinyRemapper.classes`, `.conflicts`, `.classesToMakePublic` etc

The .class file read step (`TinyRemapper.read`) populates 1.+2. depending on .class file location (both input+class path!)

The `readClasses` -> `classes` transfer step needs to do the following:
- build a `MrjState` for the base version (as before done into TinyRemapper)
- build `MrjState`s for every versioned in META-INF, populating it with copied `ClassInstance` objects. Said copy is shallow for all `ClassInstance` fields that the constructor, `init` and `addInputTags` assign, but 1-deep copied for the members. `MemberInstance` copies the constructor-assigned fields shallowly. Everything else remains at its default value

`loadMappings` and `checkClassMappings` work on "global" `ClassInstance`-independent `String`->`String` `Map`s for mappings, so they are fine to run only once for everything.

`merge` and `propagate` run for each mrj version (incl base) on their isolated `MrjState` instead of the state that `TinyRemapper` used to directly hold.

`apply` can stream the base version's `MrjState` out as-is, the other `MrjState`s need to be filtered to only yield their version-exclusive classes. The output paths need to be adjusted for META-INF.

LogicFan:
> Actually I think a 3 map approach is the easiest and still fairly quick:
> 
> 1. `Map<String, ClassInstance>` for the non-META-INF classes only (existing `readClasses`)
> 2. `Map<Integer, Map<String, ClassInstance>>` for the versioned META-INF classes only (new `readMrjClasses`)
> 3. `Map<Integer, MrjState>` for all mrj versions (base+whatever meta-inf discriminates between), `MrjState` replaces/holds `TinyRemapper.classes`, `.conflicts`, `.classesToMakePublic` etc
> 
> The .class file read step (`TinyRemapper.read`) populates 1.+2. depending on .class file location (both input+class path!)
> 
> The `readClasses` -> `classes` transfer step needs to do the following:
> 
> * build a `MrjState` for the base version (as before done into TinyRemapper) from `readClasses`
> * build `MrjState`s for every versioned in META-INF from `classes` and `readMrjClasses`, populating it with copied `ClassInstance` objects. Said copy is shallow for all `ClassInstance` fields that the constructor, `init` and `addInputTags` assign, but 1-deep copied for the members. `MemberInstance` copies the constructor-assigned fields shallowly. Everything else remains at its default value
> 
> `loadMappings` and `checkClassMappings` work on "global" `ClassInstance`-independent `String`->`String` `Map`s for mappings, so they are fine to run only once for everything.
> 
> `merge` and `propagate` run for each mrj version (incl base) on their isolated `MrjState` instead of the state that `TinyRemapper` used to directly hold.
> 
> `apply` can stream the base version's `MrjState` out as-is, the other `MrjState`s need to be filtered to only yield their version-exclusive classes. The output paths need to be adjusted for META-INF.

Yes, this would work. However, it quite different from this PR. If I change to your method, then almost everything needs to be rewritten. Hence, I would prefer the existing method. 

The existing method should work just fine from a functionality perspective. From the performance perspective, I think your biggest complain is `VersionAwareMap` (which can be rewritten when I finish the current test); anything else should be just fine.

sfPlayer1:
The problem is that you don't really have equivalents to most of these steps. As far as I can tell the isolation is not enough and it'd pollute the propagation and hierarchy impact across versions.

You are apparently clearing+reusing `classes` and maintain the extra `preprocessingClasses` instead of those `MrjState`s. `postprocessingClasses` looks to essentially be both an output container and sorting helper, but this is fairly easy to do do in apply:

```java
for (final ClassInstance cls : classes.values()) {
	if (!cls.isInput) continue;
	[...]
	futures.add(threadPool.submit(() -> immediateOutputConsumer.accept(cls, apply(cls))));
}
```
becomes
```java
for (int version : mrjStates.keySet()) {
	for (final ClassInstance cls : mrjStates.get(version).classes.values()) {
		if (!cls.isInput) continue;
		[...]
		futures.add(threadPool.submit(() -> immediateOutputConsumer.accept(cls, version, apply(cls))));
	}
}
```
and it'd just work if the shallow ClassInstance clone in my previous post always sets `isInput` to false (thus only true for classes that are an input AND have a .class with their version).

LogicFan:
> The problem is that you don't really have equivalents to most of these steps. As far as I can tell the isolation is not enough and it'd pollute the propagation and hierarchy impact across versions.
> 
> You are apparently clearing+reusing `classes` and maintain the extra `preprocessingClasses` instead of those `MrjState`s. `postprocessingClasses` looks to essentially be both an output container and sorting helper, but this is fairly easy to do do in apply:
> 
> ```java
> for (final ClassInstance cls : classes.values()) {
> 	if (!cls.isInput) continue;
> 	[...]
> 	futures.add(threadPool.submit(() -> immediateOutputConsumer.accept(cls, apply(cls))));
> }
> ```
> 
> becomes
> 
> ```java
> for (int version : mrjStates.keySet()) {
> 	for (final ClassInstance cls : mrjStates.get(version).classes.values()) {
> 		if (!cls.isInput) continue;
> 		[...]
> 		futures.add(threadPool.submit(() -> immediateOutputConsumer.accept(cls, version, apply(cls))));
> 	}
> }
> ```
> 
> and it'd just work if the shallow ClassInstance clone in my previous post always sets `isInput` to false (thus only true for classes that are an input AND have a .class with their version).

Ok, I got your point. Let me have a look how I should fix my code.

LogicFan:
After a closer look at the source, I have the following findings,

These four are only used for mappings, hence are `ClassInstance` Independent.
```
final Map<String, String> classMap = new HashMap<>();
final Map<String, String> methodMap = new HashMap<>();
final Map<String, String> methodArgMap = new HashMap<>();
final Map<String, String> fieldMap = new HashMap<>();
```

These four are only used/compared as an object reference, should be fine if two MRJ version uses different `ClassInsance` or `MemberInstance`.
```
final Set<ClassInstance> classesToMakePublic = Collections.newSetFromMap(new ConcurrentHashMap<>());
final Set<MemberInstance> membersToMakePublic = Collections.newSetFromMap(new ConcurrentHashMap<>());
private Map<ClassInstance, byte[]> outputBuffer;
final Map<MemberInstance, Set<String>> conflicts = new ConcurrentHashMap<>();
```
Specifically, the `classesToMakePublic` and `membersToMakePublic` really only care if a particular `ClassInstance` or `MemberInstance` need to be processed; and they have cleared at the end of accept function anyway. And `conflicts` only care if a particular `MemberInstance` has conflicts or not.

Thus, only `classes` need to be changed. However, I do realize that change the `classes` in the `refresh()` will not be enough (i.e. the `fixClass` function will not have the correct map), and I think the version loop should be in the `accept()` function sync block.

@sfPlayer1 Do I have any misunderstanding on the source code?

LogicFan:
Also, should I target the master branch or remapbridges branch?

LogicFan:
Additionally, since `ClassInstance` remembers `children` and `parents`, which may pollute across the version, I think it is necessary to create a new instance of `ClassInstance` for each MRJ version remapping (even the class is not specifically for this version).

sfPlayer1:
Your observations look correct, there is however a problem with `fixClass`. Since MRJ handling doesn't emit a class for every version+`ClassInstance` permutation (only those that had a corresponding .class in the first place), access widening needs to cover this gap.

If a Java 11 class A accesses an inaccessible Java 8 (base) class B, the Java 8 class B needs to be adjusted. With the version isolation `PackageAccessChecker` will mark the Java 11 `ClassInstance` B' for fixing, but not B. B' won't be output in this case, so the necessary correction won't happen.

In order to fix this `PackageAccessChecker` needs to determine the `ClassInstance` B from B', my preferred way is probably to store a reference to the original `ClassInstance` (B) in a field of the shallow copy (B'). This should also take care of complex situations like `B -> B' -> B* -> B*'` with B and B* being "real" `ClassInstance`s (B and B* resolve to themselves, B' resolves to B, B*' to B*).

Since there appears to be really only one "global" variable for MrjState, my earlier description can inline classes into it, leading to `Map<Integer, Map<String, ClassInstance>>` instead of `Map<Integer, MrjState>`+`MrjState { Map<String, ClassInstance> classes; }`. Starts to look increasingly like yours, but not quite and no guava ;)

I think working against the remapbridges branch makes much more sense, I refactored the package access check into its own class there so it wouldn't merge nicely on master and is easier to work with.

sfPlayer1:
> Additionally, since `ClassInstance` remembers `children` and `parents`, which may pollute across the version, I think it is necessary to create a new instance of `ClassInstance` for each MRJ version remapping (even the class is not specifically for this version).

This is why I do semi-shallow clones in my concept, the cloned `ClassInstance`s have their own independent hierarchy collections that way, they only share some of the fixed metadata.

LogicFan:
I shall close this PR and work on a PR against remapbridges.

Issue #9, opened 2019-05-18
Tiny V2 spec


sfPlayer1:
**this is not final**

Tiny V2 consists of a list of hierarchical sections. Every line starts a new section, whether it continues an existing
section is determined by the intendation level. A section's parent is always the closest preceding section intended
once less than itself. Accordingly a section ends just before the next line with the same or a lesser intendation level.

The child to parent relationships form the path to uniquely identify any element globally. For example all method and
field sections that are children of a class section represent members of the represented class.

Example:
```
tiny	2	0	official	intermediary	named
	someProperty	someValue
	anotherProperty
c	a	class_123	pkg/SomeClass
	m	a	(III)V	method_456	someMethod
		p	1		param_0	x
		p	2		param_1	y
		p	3		param_2	z
		c	Just a method for demonstrating the format.
	f	a	[I	field_789	someField
c	b	class_234	pkg/xy/AnotherClass
	m	a	(Ljava/lang/String;)I	method_567	anotherMethod
```

Grammar:
```
<file> ::= <header> | <header> <sections>

<header> ::= 'tiny' <tab> <major-version> <tab> <minor-version> <tab> <namespace-a> <tab> <namespace-b> <extra-namespaces> <eol> <properties>
<major-version> ::= <non-negative-int>
<minor-version> ::= <non-negative-int>
<namespace-a> ::= <namespace>
<namespace-b> ::= <namespace>
<extra-namespaces> ::= '' | <tab> <namespace> <namespaces>
<namespace> ::= <safe-string>

<properties> ::= '' | <tab> <property> <eol> <properties>
<property> ::= <property-key> | <property-key> <tab> <property-value>
<property-key> ::= <safe-string>
<property-value> ::= <escaped-string>

<sections> ::= '' | <class-section> <sections>

<class-section> ::= 'c' <tab> <class-name-a> <tab> <class-name-b> <extra-ns-class-names> <eol> <class-sub-sections>
<class-name-a> ::= <class-name>
<class-name-b> ::= <optional-class-name>
<optional-class-name> ::= '' | <class-name>
<extra-ns-cls-names> ::= '' | <tab> <optional-class-name> <extra-ns-class-names>
<class-name> ::= <conf-safe-string>
<class-sub-sections> ::= '' | <method-section> <class-sub-sections> | <field-section> <class-sub-sections> | <class-comment-section> <class-sub-sections> 

[...]

PR #54, opened 2021-05-26
MultiReleaseJar Support (re-targeting branch)

LogicFan:
Related to #53 #45. Still working in progress. Re-targeting remapbridges branch.


LogicFan:
This is my current plan to do it. Do you want to have a look @sfPlayer1 ?
rotate

LogicFan:
I would like to write a test case that MRJ actually triggers the `fixClass` method in the `TinyRemapper`. Any suggestion on how I can do that?

build.gradle
```diff
@@ -24,6 +24,8 @@ dependencies {
 	implementation 'org.ow2.asm:asm-commons:9.0'
 	implementation 'org.ow2.asm:asm-tree:9.0'
 	implementation 'org.ow2.asm:asm-util:9.0'
+
+	testImplementation 'junit:junit:4.13'
```
Juuxel, line 28: I'd use JUnit 5 instead since it's also used in other Fabric projects like Loader, Stitch, Unpick and Filament.

src/main/java/net/fabricmc/tinyremapper/ClassInstance.java
```diff
@@ -577,11 +586,35 @@ private MemberInstance getMemberPartial(MemberType type, String idPrefix) {
...
 	}
 
+	ClassInstance constructMrjCopy() {
+		// isInput should be false, since the MRJ copy should not be emitted
+		ClassInstance cls = new ClassInstance(context, false, inputTags, mrjVersion, srcPath, data);
+		cls.init(name, superName, access, interfaces);
+		// shadow copy the members and resolvedMembers, since the members should follow the ClassInstance
+		// regardless of the version
+		cls.members = members;
+		cls.resolvedMembers = resolvedMembers;
```
sfPlayer1, line 596: I meant to shallow copy the members themselves, which entails a 1-deep copy of the collection. In the case for `members` this entails leaving the members collection as-is (default empty collection), then 
```java
for (MemberInstance oldMember : origCls.getMembers()) {
	newCls.addMember(new MemberInstance(oldMember.type, newCls, oldMember.name, ...);
}
```

This ensures that MembeRInstance.cls is accurate, and the propagation-result captured by newName, etc stays in isolation.

build.gradle
LogicFan: fixed

LogicFan:
> ... requires supplying a jar+mappings, then analyzing the output jar with asm for having the right access flags.
I've written a test that can do the first part. But I am not sure what is the requirement for mapping/jar to trigger the access change. Could you provide a link on how the `.tiny` format works?

src/main/java/net/fabricmc/tinyremapper/ClassInstance.java
LogicFan: fixed

sfPlayer1:
The general problem this facility is meant to detect and/or fix is when two classes share the same package before, but not after remapping, and a access between them uses package-private or protected (=package-private+subclasses) access. So you need two .class file sharing a package and a non-public access between them. The mappings need to pull them apart / put them in separate packages, which in turn would make TR with (IIRC) --checkPackageAccess detect or with --fixPackageAccess rectify the fallout of this change by making the accessed class and/or member public.

See https://github.com/FabricMC/tiny-remapper/issues/9 for the tiny v2 format spec (the mapping file)

sfPlayer1:
For simple test cases like this (1 class with a package-private member + 1 class with a method accessing said member) it may actually be nice to generate the test classes with asm in the first place instead of putting binaries into the test resources.

LogicFan:
The test I wrote cannot be passed due to #55 .

LogicFan:
@sfPlayer1 Do I have any misunderstanding on how the fix package access work? It seems that my test is incorrect

sfPlayer1:
mapping1 shouldn't involve anything it checks since no packages hierarchies change. For anything else I first have to check, your .class files are unfortunately quite opaque. It'd be better to generate them on the fly as follows (pseudo code):
```java
ClassWriter cw = new ClassWriter(.., ClassWriter.COMPUTE_MAXS);
cw.visit("my/ClassName", Opcodes.ACC_PUBLIC, ...);
cw.visitField("someField", ...);
MethodVisitor mv = cw.visitMethod(...);
mv.visitFieldInsn(Opcodes.GET_STATIC, "my/ClassName", "someField"...);
mv.visitInsn(Opcodes.POP);
mv.isitInsn(Opcodes.RETURN);
mv.visitEnd();
cw.visitEnd();
```

to generate a simple class with a field "someField" and a method that reads the field

LogicFan:
Yes, I think to generate on the fly is more human-readable (no magic binary anymore). I will change the test.

The test used for package access is `mapping2.tiny` which involves a package change.

LogicFan:
I realize a bug in the `private static boolean isSamePackage(String clsA, int pkgEnd, String clsB)` of the `PackageAccessChecker`. 

```
clsA: "com/github/logicf/pkg1/D1"
pkgEnd: 22
clsB: "com/github/logicf/pkg2/D3"
```

would return `true`. However, I believe the expected behaviour is returning `false`. I guess this is related to #55.

Could you confirm my understanding?

LogicFan:
After some test, I can confirm that 
commit 139cd2e96aff2d392f511ae9a88ed412997bd1cc
fix the issue #55

LogicFan:
With a new test added. I am pretty confident that the code is correct. Could you do a code review for this, @sfPlayer1? Thank you.

src/main/java/net/fabricmc/tinyremapper/ClassInstance.java
```diff
@@ -147,6 +152,8 @@ public boolean isPublicOrPrivate() {
 		return (access & (Opcodes.ACC_PUBLIC | Opcodes.ACC_PRIVATE)) != 0;
 	}
 
+	public boolean isMrjCopy() { return mrjOrigin == this; }
```
sfPlayer1, line 155: shouldn't this be !=?

sfPlayer1:
I think a better fix for #55 should be `clsA.regionMatches(0, clsB, 0, pkgEnd - 1)` -> `clsA.regionMatches(0, clsB, 0, pkgEnd)`. The more complex code tries to avoid allocations and do as little as possible.

src/main/java/net/fabricmc/tinyremapper/ClassInstance.java
LogicFan: fixed

LogicFan:
I just realize that if input tag and MRJ are both present, then there are still bugs. Namely, since input tag exists, `outputBuffer` will be used, and since `outputBuffer` only set to null when `finish()` is called, on the second time, no operation will be performed as `outputBuffer == null` is false. I have a test locally that can reproduce this problem. Still thinking about a solution for this.

LogicFan:
> I just realize that if input tag and MRJ are both present, then there are still bugs. Namely, ...

This is fixed with an added test.

src/main/java/net/fabricmc/tinyremapper/TinyRemapper.java
```diff
@@ -413,7 +420,8 @@ public FileVisitResult visitFile(Path file, BasicFileAttributes attrs) throws IO
 		List<ClassInstance> ret = new ArrayList<ClassInstance>();
 
 		if (file.toString().endsWith(".class")) {
-			ClassInstance res = analyze(isInput, tags, srcPath, Files.readAllBytes(file));
+			// flat class file cannot have MRJ, just set it to default version
```
sfPlayer1, line 423: I believe they can, imagine e.g. Eclipse bin dirs that contains the equivalent of the Jar contents, which may as well be `/path/to/bin/META-INF/classes/.../Bla.class`

src/main/java/net/fabricmc/tinyremapper/TinyRemapper.java
```diff
@@ -744,71 +756,102 @@ public void apply(final BiConsumer<String, byte[]> outputConsumer, InputTag... i
...
-					outputBuffer = new ConcurrentHashMap<>();
-					immediateOutputConsumer = outputBuffer::put;
-				} else {
-					immediateOutputConsumer = (cls, data) -> outputConsumer.accept(mapClass(cls.getName()), data);
-				}
+			for (int version: mrjClasses.keySet()) {
+				mrjRefresh(version);
 
-				List<Future<?>> futures = new ArrayList<>();
+				if (outputBuffer == null) { // first (inputTags present) or full (no input tags) output invocation, process everything but don't output if input tags are present
```
sfPlayer1, line 762: Moving outputBuffer inside the version loop doesn't make immediate sense to me. It is meant to facilitate post processing needed by package access fixing (changing the output after processing everything) or to make tagged input handling faster (input tags are for multiple inputs, e.g. jarA and jarB, that are both on the class path but need to be output separately, e.g. remapped jarA and remapped jarB, not a single jar containing the contents of A+B).

src/main/java/net/fabricmc/tinyremapper/TinyRemapper.java
```diff
@@ -744,71 +756,102 @@ public void apply(final BiConsumer<String, byte[]> outputConsumer, InputTag... i
...
+
+			if(mrjClasses.put(toVersion, toClasses) != null) {
+				throw new RuntimeException("internal error: duplicate versions in mrjClasses");
+			}
 
-				if (inputTags == null || cls.hasAnyInputTag(inputTags)) {
-					outputConsumer.accept(mapClass(cls.getName()), entry.getValue());
+			// find the fromVersion that just lower the the toVersion
+			Optional<Integer> fromVersion = mrjClasses.keySet().stream()
+					.filter(v -> v < toVersion).max(Integer::compare);
```
sfPlayer1, line 850: I think you can do that cleaner by rememberint the toVersion of the previous version loop iteration.

sfPlayer1 (commented):
Looks quite good already, I left a few notes inline regarding some specific details.

src/main/java/net/fabricmc/tinyremapper/TinyRemapper.java
LogicFan: I see, then I guess we need to check if the path contains `META-INF/classes` and process it.

src/main/java/net/fabricmc/tinyremapper/TinyRemapper.java
LogicFan: This is intended to fix a bug when MRJ and input tag are both presents. In this case, if the output buffer is outside of the loop, and `outputBuffer` will be cleared for each mrj version, so only the last-processing mrj class file will be outputted. 

If I do not clear outputBuffer for each mrj version, then the following would happen.
> Namely, since input tag exists, outputBuffer will be used, and since outputBuffer only set to null when finish() is called, on the second time, no operation will be performed as outputBuffer == null is false. I have a test locally that can reproduce this problem. Still thinking about a solution for this.

It is possible some better solution exists for this problem. But this is the one that first came into my head.

LogicFan: By the way, the `mrj4` test in the `IntegrationTest1` is specifically for this. If you try to move the `outputBuffer` inside the loop and see what happens.

src/main/java/net/fabricmc/tinyremapper/TinyRemapper.java
LogicFan: Emm. I do not see how that would be cleaner, and I don't understand how you want to change it. Maybe you could change the code on GitHub. As long as the functionalities are the same, I am fine with it.

src/main/java/net/fabricmc/tinyremapper/TinyRemapper.java
LogicFan: fixed

src/main/java/net/fabricmc/tinyremapper/TinyRemapper.java
sfPlayer1: I tried to PR to your branch, but GH failed. I meant like so, assuming its logic is accurate:

```java
private void fixMrjClasses(Set<Integer> versions) {
	int lastVersion = 0;

	// ensure the new version is added from lowest to highest
	for (int toVersion: versions.stream().sorted().collect(Collectors.toList())) {
		Map<String, ClassInstance> toClasses = new HashMap<>();

		if (mrjClasses.put(toVersion, toClasses) != null) {
			throw new RuntimeException("internal error: duplicate versions in mrjClasses");
		}

		if (lastVersion > 0) {
			Map<String, ClassInstance> fromClasses = mrjClasses.get(lastVersion);

			for (ClassInstance cls: fromClasses.values()) {
				addClass(cls.constructMrjCopy(), toClasses, false);
			}
		}

		lastVersion = toVersion;
	}
}
```

LogicFan: This code may not work properly. Suppose `mrjClasses` have version `8` and `9`, and `versions` contain `11`. Then the loop will only run once and therefore old classes won't be added to the `mrjClasses`.
The difference is that your code is trying to find an appreciated version from `toVersion` (i.e. `versions`), where my code is trying to find appreciate version from `mrjClasses`. The latter one will contain more versions than `toVersion`.

sfPlayer1: You mean like when letting TR run, add another input, then run it again?

LogicFan: yes.

src/main/java/net/fabricmc/tinyremapper/TinyRemapper.java
```diff
@@ -408,12 +418,32 @@ public FileVisitResult visitFile(Path file, BasicFileAttributes attrs) throws IO
...
+		}
+	}
+
+	private static int getMrjVersionFromPath(String str) {
+		Pattern pattern = Pattern.compile("(?<=" + ClassInstance.MRJ_PREFIX + "/)[0-9]*");
+		Matcher matcher = pattern.matcher(str);
+		if (matcher.find()) {
+			return Integer.parseInt(matcher.group());
+		}
+		return ClassInstance.MRJ_DEFAULT;
```
sfPlayer1, line 437: I don't really like how this just searches the path string, it should be possible to determine this accurately due to the class name <-> path relationship in Java.

I'd do the following:
- instead of passing `int mrjVersion` and `byte[] data` to `analyze()`, pass the `file` and whether it's a .jar or .class
- do the `Files.readAllBytes` in `analyze()`
- set the mrjVersion in `ClassInstance.init()` instead of the constructor
- determine said mrjVersion just before the `ClassInstance.init()` call in `visit()` using the class name and `file`

The last step works as follows for plain classes:
1. ensure the file path is absolute and normalized by making it so all the way back in `net.fabricmc.tinyremapper.TinyRemapper.read(Path, boolean, InputTag[], Path, boolean, List<FileSystem>)`
2. strip the class name parts from `file`, so `/path/to/some/pkg/TheClass.class` for `some.okg.TheClass` becomes `/path/to`
3. the parent dir of the path is now either null, the mrj version or something else
4. the parent dir of the parent dir is now versions for mrj, and the parent of that META-INF

For jars replace 2.-4. with checking the 1st component of the path for META-INF and the 2nd for the versions, the 3rd is the actual version

src/main/java/net/fabricmc/tinyremapper/TinyRemapper.java
sfPlayer1: Good to know, not sure it is safe to not start over with the mrj copies in every case though, which is why I didn't expect this.

sfPlayer1:
Btw since we're approaching the merge please make sure the wildcard imports are gone and spaces are converted to tabs.

src/main/java/net/fabricmc/tinyremapper/TinyRemapper.java
LogicFan: Just realize the current code still have the possibility to have the wrong MRJ resolution. Let me fix that.

src/main/java/net/fabricmc/tinyremapper/TinyRemapper.java
LogicFan: fixed

src/main/java/net/fabricmc/tinyremapper/TinyRemapper.java
LogicFan: fixed

sfPlayer1 (approved):
I think it's good to merge now, thanks for all the work! Further changes can be done in followup PRs/commits.

PR #56, opened 2021-05-29
add resource remapper

Devan-Kerman:
allows mods to add additional remapping to resources (eg. access wideners), still a WIP. Perhaps I should move the META-INF remapping to this API, would also let me know what additional context I should pass


Devan-Kerman:
not 100% sure if the behavior is exactly the same, but I moved the META-INF fix/skip code into the resource remappers

Devan-Kerman:
superseded by https://github.com/FabricMC/tiny-remapper/pull/57

PR #58, opened 2021-05-30
add annotation visitor wrapping

Devan-Kerman:
Allows for the creation of a wrapper around annotation visitors, this can be used for annotation remapping.


Devan-Kerman:
I think this needs a restart, will open a new PR

PR #62, opened 2021-06-12
Pre & Post ClassVisitor

Devan-Kerman:
a pre & post-class visitor, along with a small API


Devan-Kerman:
need to fix some compile errors I think, merge probably wasn't perfect

modmuss50:
What is the usecase for this?

Devan-Kerman:
remapping mixins and other bits

modmuss50:
Oh i see, for somereason I throught this was the access widener repo, opps...

This is one for @sfPlayer1 to look at then.

src/main/java/net/fabricmc/tinyremapper/ClassInstance.java
```diff
@@ -652,6 +705,7 @@ public static String getMrjName(String clsName, int mrjVersion) {
 	private String name;
 	private int mrjVersion;
 	private String superName;
+	private String signature;
```
LogicFan, line 708: What is the difference between signature and name?

src/main/java/net/fabricmc/tinyremapper/TinyRemapper.java
```diff
@@ -161,14 +164,24 @@ public Builder extraRemapper(Remapper remapper) {
 			return this;
 		}
 
+		public Builder extraPreVisitor(WrapperFunction func) {
+			this.pre = func;
+			return this;
+		}
+
+		public Builder extraPostVisitor(WrapperFunction function) {
```
LogicFan, line 172: parameter name should be consistent.

LogicFan:
I am interested in using this new feature to add to the ability to remap mixin inside tiny-remapper. Do you have any idea how that can be done?

src/main/java/net/fabricmc/tinyremapper/api/WrapperFunction.java
```diff
@@ -0,0 +1,7 @@
+package net.fabricmc.tinyremapper.api;
+
+import org.objectweb.asm.ClassVisitor;
+
+public interface WrapperFunction {
```
LogicFan, line 5: Maybe add documentation on how this one should be used.

src/main/java/net/fabricmc/tinyremapper/TinyRemapper.java
```diff
@@ -206,11 +220,13 @@ private TinyRemapper(Collection<IMappingProvider> mappingProviders, boolean igno
...
 			boolean skipLocalMapping,
 			boolean renameInvalidLocals,
-			ClassVisitor extraAnalyzeVisitor, Remapper extraRemapper) {
+			ClassVisitor extraAnalyzeVisitor, Remapper extraRemapper, WrapperFunction pre, WrapperFunction post) {
 		this.mappingProviders = mappingProviders;
 		this.ignoreFieldDesc = ignoreFieldDesc;
 		this.threadCount = threadCount > 0 ? threadCount : Math.max(Runtime.getRuntime().availableProcessors(), 2);
 		this.keepInputData = keepInputData;
+		this.pre = pre;
+		this.post = post;
```
LogicFan, line 229: Why would you only allow one pre/post visitor. Imagine that there could be different pre/post visitors for different purposes and written by different people, they should all be able to apply in sequence. For example, it should allow `builder.extraPre(...).extraPre(...).extraPre(...)` and apply them in sequence.

Chocohead: `ClassVisitor`s naturally chain, so if you had multiple you wanted to apply you would simply chain them in your `WrapperFunction`

src/main/java/net/fabricmc/tinyremapper/ClassInstance.java
Chocohead: The name is the internal name of the class, signature is the generic signature of the class (which may be `null` if the class isn't parametrised)

src/main/java/net/fabricmc/tinyremapper/AsmClassRemapper.java
```diff
@@ -113,17 +115,19 @@ public MethodVisitor visitMethod(int access, String name, String descriptor, Str
 			methodNode = new MethodNode(api, access, name, descriptor, signature, exceptions);
 		}
 
-		return super.visitMethod(access, name, descriptor, signature, exceptions);
+		MethodVisitor mv = super.visitMethod(access, name, descriptor, signature, exceptions);
+		MemberHeader header = null;
```
Chocohead, line 119: Should this be `null`?

src/main/java/net/fabricmc/tinyremapper/AsmClassRemapper.java
```diff
@@ -96,11 +96,13 @@ public void visitSource(String source, String debug) {
 
 	@Override
 	public FieldVisitor visitField(int access, String name, String descriptor, String signature, Object value) {
+		MemberHeader header = null;
```
Chocohead, line 99: Should this be `null`?

build.gradle
```diff
@@ -18,6 +18,10 @@ archivesBaseName = 'tiny-remapper'
 
 repositories {
 	mavenCentral()
+	maven {
+		name = 'fabric'
+		url = 'https://maven.fabricmc.net/'
+	}
```
Chocohead, line 24: Suppose we don't need this now Mixin isn't a dependency

src/main/java/net/fabricmc/tinyremapper/ClassInstance.java
```diff
@@ -148,10 +163,48 @@ public int getMrjVersion() {
...
 	}
 
+	@Override
+	public String getSignature() {
+		return signature;
+	}
+
+	@Override
+	public MemberHeader getField(String name, String desc) {
+		MemberInstance instance = this.resolvedMembers.get(MemberInstance.getFieldId(name, desc, false));
```
Chocohead, line 178: Could allow `null` field descriptors if `false` is switched to `desc == null`

src/main/java/net/fabricmc/tinyremapper/TinyRemapper.java
LogicFan: But then the function name should not be called `extraPre` or `extraPost` as it confuses people. It should be `setPre` or `setPost` since the "extra" indicates you can have them chain together.

Devan-Kerman: the other methods don't follow that convention

LogicFan (approved):

src/main/java/net/fabricmc/tinyremapper/api/WrapperFunction.java
LogicFan: Maybe also consider adding `@FunctionalInterface` for this interface.

src/main/java/net/fabricmc/tinyremapper/api/WrapperFunction.java
```diff
@@ -0,0 +1,10 @@
+package net.fabricmc.tinyremapper.api;
+
+import org.objectweb.asm.ClassVisitor;
+
+/**
+ * @see ClassVisitor#ClassVisitor(int, ClassVisitor)
+ */
+public interface WrapperFunction {
+	ClassVisitor wrap(ClassVisitor visitor, ExtendedRemapper remapper, Classpath classpath);
```
LogicFan, line 9: Maybe also consider adding @FunctionalInterface for this interface.

src/main/java/net/fabricmc/tinyremapper/TinyRemapper.java
```diff
@@ -956,11 +971,21 @@ private void mrjRefresh(MrjState state) {
...
-		reader.accept(new AsmClassRemapper(visitor, cls.getContext().remapper, rebuildSourceFilenames, checkPackageAccess, skipLocalMapping, renameInvalidLocals), flags);
+		if (post != null) {
+			visitor = post.wrap(visitor, this.getRemapper(), this);
+		}
+
+		visitor = new AsmClassRemapper(visitor, cls.getContext().remapper, rebuildSourceFilenames, checkPackageAccess, skipLocalMapping, renameInvalidLocals);
+
+		if (pre != null) {
+			visitor = pre.wrap(visitor, this.getRemapper(), this);
+		}
```
LogicFan, line 985: the `this.getRemapper()` does not respect with the mrj version. Shouldn't this be `cls.getContext.remapper`?

Chocohead: This PR was made before the MRJ PR was merged so this is probably a merging oversight

Chocohead:
Are there any changes to how `AsmClassRemapper` works now? The only remaining changes look either to be stylistic or functionally no different to how it previously worked (`createMethodRemapper` and `createFieldRemapper`)

src/main/java/net/fabricmc/tinyremapper/MetaInfFixer.java
```diff
@@ -39,7 +39,7 @@ public void transform(Path destinationDirectory, Path relativePath, InputStream
 			Path outputDir = outputFile.getParent();
 			if (outputDir != null) Files.createDirectories(outputDir);
 
-			try (OutputStream os = new BufferedOutputStream(Files.newOutputStream(outputFile))) {
+			try (OutputStream os = new BufferedOutputStream(Files.newOutputStream(outputFile, StandardOpenOption.CREATE))) {
```
Chocohead, line 42: The default options (having passed none) is `StandardOpenOption.CREATE, StandardOpenOption.TRUNCATE_EXISTING, StandardOpenOption.WRITE`, is this for something like avoiding overwriting service files?

Devan-Kerman: I think that's how it was originally, maybe it was a bad merge

src/main/java/net/fabricmc/tinyremapper/api/MemberHeader.java
```diff
@@ -0,0 +1,44 @@
+package net.fabricmc.tinyremapper.api;
+
+import java.util.Objects;
+
+public class MemberHeader {
+	public final ClassHeader header;
```
liach, line 6: Probably call this field `owner` instead. `header.header` is quite confusing

src/main/java/net/fabricmc/tinyremapper/api/WrapperFunction.java
liach: Must the visitor propagate at real time or can it cache the results and mass dump calls later? Such as we first grab everything from the source visitor to a class node, do stuff to the node, and finally we dump a visitor to the node, instead of delegating most calls and only patching a few selected calls.

src/main/java/net/fabricmc/tinyremapper/api/WrapperFunction.java
liach: Should add a simple API like `andThen` to chain wrappers easily, and use that to support multiple wrappers in remapper builder code

liach:
2 comments on `WrapperFunction`:
1. We can just add an `andThen` default method (or static method) to chain pre/post wrappers in the remapper builder
2. For wrapping usage:
Must the visitor propagate at real time or can it cache the results and mass dump calls later? Such as we first grab everything from the source visitor to a class node, do stuff to the node, and finally we dump a visitor to the node, instead of patching a few selected calls and delegating other calls. If this behavior is prohibited, it may be worthy to document.

Thanks github, my 2 comments on desired lines are considered followup comments of the resolved discussion. :vomiting_face:

LogicFan:
> Must the visitor propagate at real time or can it cache the results and mass dump calls later?

Both can be done with the current ASM structure. There is no reason to ban one of them, and both of them have use cases.

liach:
Yeah, that should be fine, but if there is any peculiarities caused by such buffering, I mean it may be worthy to document.

LogicFan:
I think it is also necessary to have a visitor when reads the class; or have the ability to work by round globally. For example, if I want to remap a public `@Shadow` field, it would be necessary to remap the whole thing instead of just a single class. The current API lacks this functionality.

LogicFan:
Why would you want to feed `Classpath` as a parameter? You can get it inside `ClassVisitor`, right?

LogicFan:
Any ETA when this one will be merged? #63 depends on this.

Devan-Kerman:
I think I will add an andThen simply for convienence

liach:
@sfPlayer1 mind review?

src/main/java/net/fabricmc/tinyremapper/ClassInstance.java
```diff
@@ -148,10 +163,48 @@ public int getMrjVersion() {
...
+			return new MemberHeader(this, instance.access, instance.name, instance.desc);
+		}
+
+		return null;
+	}
+
+	@Override
+	public Iterable<MemberHeader> allMembers() {
+		return MappedIterator.map(this.resolvedMembers.values(), i -> new MemberHeader(this, i.access, i.name, i.desc));
+	}
```
LogicFan, line 201: `resolvedMembers` is a cache. It may or may or contains the member from parents' classes, which introduces inconsistency. I think a better way is needed.

src/main/java/net/fabricmc/tinyremapper/api/MemberHeader.java
```diff
@@ -0,0 +1,44 @@
+package net.fabricmc.tinyremapper.api;
+
+import java.util.Objects;
+
+public class MemberHeader {
```
LogicFan, line 5: Is there any specific reason to not make this an interface inherited by `MemberInstance`?

LogicFan (changes requested):

LogicFan:
Here is how `resolve` is written in the `ClassInstance`.

```Java
public MemberInstance resolve(MemberType type, String id) {
		MemberInstance member = getMember(type, id);
		if (member != null) return member;

		// get from cache
		member = resolvedMembers.get(id);

		if (member == null) {
			// compute
			member = resolve0(type, id);
			assert member != null;

			// put in cache
			MemberInstance prev = resolvedMembers.putIfAbsent(id, member);
			if (prev != null) member = prev;
		}

		return member != nullMember ? member : null;
	}
```

You can see that `resolvedMemebrs` is a cache that generated when necessary. i.e. the member resolution is lazy. Thus, you cannot directly use `resolvedMembers`.

LogicFan:
My suggestion is to make the `MemberHeader` an interface inherited by `MemberInstance`, and use `resolve` to get the method & field for `getMethod` and `getField`. 

`allMembers` is tricky, either you need to recursively resolve all members from parents, or just remove this method I guess.

LogicFan:
Any update on this PR?

Devan-Kerman:
I was out, am back now, will fix

Devan-Kerman:
everything should be fixed now

src/main/java/net/fabricmc/tinyremapper/ClassInstance.java
```diff
@@ -148,10 +163,37 @@ public int getMrjVersion() {
...
+
+	@Override
+	public MemberHeader getMethod(String name, String desc) {
+		return this.resolve(MemberType.METHOD, MemberInstance.getMethodId(name, desc));
+	}
+
+	@Override
+	@SuppressWarnings("ALL")
+	public Iterable<MemberHeader> allMembers() {
+		return (Iterable) this.members.values();
```
Chocohead, line 189: You could fix the nasty generics here doing `Iterable<? extends MemberHeader>` as the return type

Devan-Kerman: yea but then the api looks uglier, better to have the ugly in impl than in API

Chocohead:
> 
> 
> Are there any changes to how `AsmClassRemapper` works now? The only remaining changes look either to be stylistic or functionally no different to how it previously worked (`createMethodRemapper` and `createFieldRemapper`)

Think this question got missed in between reviews too

src/main/java/net/fabricmc/tinyremapper/AsmClassRemapper.java
```diff
@@ -100,7 +99,7 @@ public FieldVisitor visitField(int access, String name, String descriptor, Strin
 			PackageAccessChecker.checkDesc(className, descriptor, "field descriptor", (AsmRemapper) remapper);
 		}
 
-		return super.visitField(access, name, descriptor, signature, value);
+		return new AsmFieldRemapper(super.visitField(access, name, descriptor, signature, value), remapper);
 	}
```
LogicFan, line 103: Any reason to change the remap logic here? (And some other places in the AsmClassRemapper).

src/main/java/net/fabricmc/tinyremapper/ClassInstance.java
Chocohead: It's an `Iterable` though, your normal use in a `for` loop wouldn't be any different. If it was returning `Collection` directly then I'd understand, as it would as it would stop you adding to it (not that you'd be able to in this particular instance anyway), but here it is just using generics as they were intended

LogicFan: There are two things I want to comment,
1. I really don't think `allMembers` is a good name since it does not retrieve all members, it only retrieves the members defined in this class, not its interface or superclass.
2. Why use iterable, not collection. I personally think the collection is easier to work with than iterable.

sfPlayer1:
What about renaming *Header to Tr*?

src/main/java/net/fabricmc/tinyremapper/ClassInstance.java
```diff
@@ -148,10 +163,37 @@ public int getMrjVersion() {
...
+
+	@Override
+	public MemberHeader getField(String name, String desc) {
+		return this.resolve(MemberType.FIELD, MemberInstance.getFieldId(name, desc, desc == null));
+	}
+
+	@Override
+	public MemberHeader getMethod(String name, String desc) {
+		return this.resolve(MemberType.METHOD, MemberInstance.getMethodId(name, desc));
+	}
```
LogicFan, line 184: Can you also have a version of `getField` and `getMethod` that does not require desc? Because my implementation in #63 would require this capability. You can use `resolvePartial` for this purpose.

LogicFan (changes requested):

LogicFan:
> What about renaming `*Header` to `Tr*`?

This will break lots of code in #63 as it starts uses this API already. Any specific reason you want to make this change?

sfPlayer1:
> > What about renaming `*Header` to `Tr*`?
> 
> This will break lots of code in #63 as it starts uses this API already. Any specific reason you want to make this change?

The Header suffix makes no sense to me, better to use Tr (for tiny remapper) for such a special purpose API. It may be quite easy to merge into your PR if you squash everything, do the same refactors as here, then rebase onto the merged #62.

src/main/java/net/fabricmc/tinyremapper/api/ClassHeader.java
```diff
@@ -0,0 +1,30 @@
...
+	/**
+	 * get field by name and descriptor, this is not cached.
+	 * @param desc can be null
+	 */
+	MemberHeader getField(String name, String desc);
+
+	/**
+	 * get method by name and descriptor, this is not cached.
+	 */
+	MemberHeader getMethod(String name, String desc);
```
LogicFan, line 27: Since you use `resolve`, which uses the exact match. The documentation here is not correct anymore.

src/main/java/net/fabricmc/tinyremapper/ClassInstance.java
LogicFan: >  your normal use in a for loop wouldn't be any different.

I usually work with `Stream`. Can `Iterable` easily converted to stream? Sorry for the stupid question, I really never touched iterable before.

src/main/java/net/fabricmc/tinyremapper/api/WrapperFunction.java
```diff
@@ -0,0 +1,33 @@
...
+
+			return current;
+		};
+	}
+
+	ClassVisitor wrap(ClassVisitor visitor, ExtendedRemapper remapper, Classpath classpath);
+
+	default WrapperFunction andThen(WrapperFunction target) {
+		return (visitor, remapper, classpath) -> target.wrap(this.wrap(visitor, remapper, classpath), remapper, classpath);
+	}
```
sfPlayer1, line 32: Is there any point to the static methods? The API user shouldn't be encouraged to chain wrap or otherwise split the visitors.

LogicFan:
Here is a reference implementation for `getMethod` and `getField` that does not use desc that I used in my PR

```Java
	@Override
	public MemberHeader getField(String name) {
		return this.resolvePartial(MemberType.FIELD, name, ";;");
	}
	@Override
	public MemberHeader getMethod(String name) {
		return this.resolvePartial(MemberType.METHOD, name, "(");
	}
```

src/main/java/net/fabricmc/tinyremapper/api/Classpath.java
```diff
@@ -0,0 +1,10 @@
+package net.fabricmc.tinyremapper.api;
+
+public interface Classpath {
```
sfPlayer1, line 3: The name of this class is a bit weird considering what it does. I call this Environment in Matcher, so maybe `TrEnvironment`? getByName would have to be called getClass or similar then.

src/main/java/net/fabricmc/tinyremapper/api/ClassHeader.java
```diff
@@ -0,0 +1,30 @@
...
+
+	int getAccess();
+
+	String getName();
+
+	String getSuperName();
+
+	String getSignature();
+
+	List<String> getInterfaceList();
```
sfPlayer1, line 16: getInterfaces? Why is this a List and the other is an Iterable?

src/main/java/net/fabricmc/tinyremapper/api/ClassHeader.java
```diff
@@ -0,0 +1,30 @@
...
+	 * @param desc can be null
+	 */
+	MemberHeader getField(String name, String desc);
+
+	/**
+	 * get method by name and descriptor, this is not cached.
+	 */
+	MemberHeader getMethod(String name, String desc);
+
+	Iterable<MemberHeader> allMembers();
```
sfPlayer1, line 29: Iterable is usually a bad type to export, Collection is much better.

LogicFan:
> The Header suffix makes no sense to me, better to use Tr (for tiny remapper) for such a special purpose API. It may be quite easy to merge into your PR if you squash everything, do the same refactors as here, then rebase onto the merged #62.

Ya, that would make sense.

src/main/java/net/fabricmc/tinyremapper/TinyRemapper.java
```diff
@@ -1181,7 +1215,12 @@ public void run() {
 		/**
 		 * Propagate names into methods and create additional bridges to keep the normally mapped method name intact.
 		 */
-		COMPATIBLE;
+		COMPATIBLE
+	}
+
+	@Override
+	public ClassHeader getByName(String internalName) {
+		return this.defaultState.classes.get(internalName);
```
sfPlayer1, line 1223: This isn't really MRJ compatible, apply is MRJ aware, so the environment may be better served by MrjState?.

src/main/java/net/fabricmc/tinyremapper/TinyRemapper.java
```diff
@@ -1142,17 +1178,15 @@ public void run() {
...
-					MemberInstance getter = cls.getMember(MemberType.METHOD, getterIdSrc);
+					MemberInstance getter = cls.getMember(MemberHeader.MemberType.METHOD, getterIdSrc);
 
 					if (getter != null && getter.isVirtual()) {
 						visitedUp.add(cls);
 						visitedDown.add(cls);
-						cls.propagate(TinyRemapper.this, MemberType.METHOD, className, getterIdSrc, nameDst,
-								Direction.ANY, true, true,
-								true, visitedUp, visitedDown);
+						cls.propagate(TinyRemapper.this, MemberHeader.MemberType.METHOD, className, getterIdSrc, nameDst, Direction.ANY, true, true, true, visitedUp, visitedDown);
```
sfPlayer1, line 1189: Please avoid unrelated changes, the MemberType changes above can be eliminated by changing the import as well.

src/main/java/net/fabricmc/tinyremapper/MappedIterator.java
```diff
@@ -0,0 +1,28 @@
+package net.fabricmc.tinyremapper;
+
+import java.util.Iterator;
+import java.util.function.Function;
+
+public final class MappedIterator<F, T> implements Iterator<T> {
```
sfPlayer1, line 6: What is the point of this? Seems unused?

src/main/java/net/fabricmc/tinyremapper/ClassInstance.java
sfPlayer1: I'd use `Collection<? extends TrMember> getMembers()`

src/main/java/net/fabricmc/tinyremapper/ClassInstance.java
sfPlayer1: I'd call resolving lookup methods resolveField/resolveMethod, compared to getField/getMethod they search super classes/interfaces for the member - not just the class itself.

Devan-Kerman:
There are no intentional changes to AsmRemapper

Devan-Kerman:
Prototype is an alternative name

sfPlayer1:
IMO the overly generic sounding/universal API names should be Tr prefixed, this is a very special purpose library that sits in every mod dev's class path and thus risks messing with auto-imports.

Devan-Kerman:
fair enough

src/main/java/net/fabricmc/tinyremapper/api/ClassHeader.java
Devan-Kerman: Order is important here but not for members (I think)

Devan-Kerman:
does TR* work instead of Tr*? TR is an acronym

src/main/java/net/fabricmc/tinyremapper/MappedIterator.java
Devan-Kerman: it's unused now yea, forgot to remove it

src/main/java/net/fabricmc/tinyremapper/TinyRemapper.java
Devan-Kerman: this fails checkstyle

sfPlayer1:
abbreviations always count as a single word (Url, Http, Tr, ...)

src/main/java/net/fabricmc/tinyremapper/AsmClassRemapper.java
Devan-Kerman: just passing the remapper instance

Devan-Kerman: just passing the remapper instance

src/main/java/net/fabricmc/tinyremapper/ClassInstance.java
Devan-Kerman: might be better to make desc nullable

src/main/java/net/fabricmc/tinyremapper/ClassInstance.java
```diff
@@ -174,19 +174,22 @@ public String getSignature() {
 	}
 
 	@Override
-	public MemberHeader getField(String name, String desc) {
+	public TrMember resolveField(String name, String desc) {
 		return this.resolve(MemberType.FIELD, MemberInstance.getFieldId(name, desc, desc == null));
 	}
```
LogicFan: The internal implementation uses `tr.ignoreFieldDesc` everywhere. Say if the `tr.ignoreFieldDesc = true` and someone passes `null`, then it will always return no match since the internal id does not match (one id has desc another id does not). Vice versa.

LogicFan: Here is code that make things work

```Java
	@Override
	public TrMember resolveField(String name, String desc) {
		if (desc == null) {
			return this.resolvePartial(MemberType.FIELD, name, ";;");
		} else {
			return this.resolve(MemberType.FIELD, MemberInstance.getFieldId(name, desc, tr.ignoreFieldDesc));
		}
	}
```

LogicFan (approved):
I think this PR is good to go. My only concern is the change of `AsmClassRemapper`, which is the part of the code I do not fully understand. @sfPlayer1 mind to have a look again?

src/main/java/net/fabricmc/tinyremapper/AsmClassRemapper.java
sfPlayer1: I don't get it, the remapper is being passed by the previous implementation as well? It just funneled the remapper instance through the super class field

src/main/java/net/fabricmc/tinyremapper/api/TrClass.java
```diff
@@ -0,0 +1,29 @@
...
+
+	int getAccess();
+
+	String getName();
+
+	String getSuperName();
+
+	String getSignature();
+
+	List<String> getInterfaceList();
```
sfPlayer1, line 17: getInterfaces

src/main/java/net/fabricmc/tinyremapper/api/TrClass.java
```diff
@@ -0,0 +1,29 @@
...
+	/**
+	 * @param desc can be null
+	 */
+	TrMember resolveField(String name, String desc);
+
+	/**
+	 */
+	TrMember resolveMethod(String name, String desc);
+
+	Collection<? extends TrMember> allMembers();
```
sfPlayer1, line 28: getMembers

src/main/java/net/fabricmc/tinyremapper/api/TrEnvironment.java
```diff
@@ -0,0 +1,10 @@
+package net.fabricmc.tinyremapper.api;
+
+public interface TrEnvironment {
+	/**
+	 * @return the class with the passed name, or null if not found.
+	 */
+	TrClass getByName(String internalName);
```
sfPlayer1, line 7: getClass

src/main/java/net/fabricmc/tinyremapper/api/TrClass.java
LogicFan: I think it is also better to change the return type of this method to a Collection since the `getMembers` is Collection.

LogicFan:
Any update for this PR?

Devan-Kerman:
should be updated now

sfPlayer1 (approved):
LGTM now, thanks!

Issue #47, opened 2021-02-18
Unable to remap class if two parents contain methods with same mapped name but different intermediary names


CoolMineman:
If a class extending `Item` implements `ToolMaterial` and overrides getEnchantability it creates the following error when compiling in loom. This is because `Item`'s getEnchantability is method_7837 in yarn and `ToolMaterial`'s getEnchantability method_8026. Maybe tiny-remapper should dupe the method and warn?
```
:remapping fluxtech-2.0.0-dev.jar
Mapping source name conflicts detected:
  net/minecraft/item/ArmorItem METHOD getEnchantability (()I) -> [net/minecraft/item/Item/method_7837, net/minecraft/item/ArmorItem/method_8026]
  net/minecraft/item/TridentItem METHOD getEnchantability (()I) -> [net/minecraft/item/Item/method_7837, net/minecraft/item/TridentItem/method_8026]
  net/minecraft/item/ToolMaterial METHOD getEnchantability (()I) -> [net/minecraft/item/Item/method_7837, net/minecraft/item/ToolMaterial/method_8026]
  net/minecraft/item/ToolItem METHOD getEnchantability (()I) -> [net/minecraft/item/Item/method_7837, net/minecraft/item/ToolItem/method_8026]
  net/minecraft/item/RangedWeaponItem METHOD getEnchantability (()I) -> [net/minecraft/item/Item/method_7837, net/minecraft/item/RangedWeaponItem/method_8026]
  net/minecraft/item/ToolMaterials METHOD getEnchantability (()I) -> [net/minecraft/item/Item/method_7837, net/minecraft/item/ToolMaterials/method_8026]
  com/fusionflux/fluxtech/items/SwordTestItem METHOD getEnchantability (()I) -> [net/minecraft/item/Item/method_7837, com/fusionflux/fluxtech/items/SwordTestItem/method_8026]
  net/minecraft/item/BookItem METHOD getEnchantability (()I) -> [net/minecraft/item/Item/method_7837, net/minecraft/item/BookItem/method_8026]
  net/minecraft/item/Item METHOD getEnchantability (()I) -> [net/minecraft/item/Item/method_7837, net/minecraft/item/Item/method_8026]
  net/minecraft/item/FishingRodItem METHOD getEnchantability (()I) -> [net/minecraft/item/Item/method_7837, net/minecraft/item/FishingRodItem/method_8026]
```

liach:
Related: https://github.com/FabricMC/yarn/pull/2601

PR #65, opened 2021-07-09
Attempted fix for #47

Devan-Kerman:
Dunno if it actually works, might fix https://github.com/FabricMC/tiny-remapper/issues/47


Devan-Kerman:
targetted wrong branch, bru

Chocohead:
Remember the inverse case of merging the multiple methods back into one, it's VM dependent if it lets you have duplicate methods in a class

Devan-Kerman:
Yea, I will prolly redo this PR, this approach is actually terrible anyways. Copying the entire method is a terrible solution, I should instead just call the other method

PR #66, opened 2021-07-18
Pre & post class visitor api

LogicFan:
This is basically #62 . Since the original author does not respond to the review very frequently, I will start this PR from where he lefted. Hope this can speed up the progress.


src/main/java/net/fabricmc/tinyremapper/api/ExtendedRemapper.java
```diff
@@ -0,0 +1,20 @@
+package net.fabricmc.tinyremapper.api;
+
+import org.objectweb.asm.commons.Remapper;
+
+public abstract class ExtendedRemapper extends Remapper {
```
LogicFan, line 5: Should this be called TrRemapper to maintain consistency?

LogicFan:
@sfPlayer1 , do you mind having a look at this?

LogicFan:
Closed since #62 is merged.

PR #67, opened 2021-07-19
fix #64

### LINKED ISSUE #64: The current `extraAnalyzeVisitor` API is hard to use.
Because of the way that the API is designed, there is a single `ClassVisitor` instance for processing all classes. Additionally, notice that this `ClassVisitor` is called in a concurrent environment, which makes the thing worse. In order to pass the value across `@Override` methods (e.g. `visit***`), the only way is to use a thread-local variable and remove it at the `visitEnd`. This is way too hacky and easy for developers to mess up.

Thus, I propose to have an overload of `extraAnalyzeVisitor` that takes function wrapper similar to #62 that creats `ClassVisitor` on the fly. 

To maintain backward compatibility, the old method that takes a `ClassVisitor` will still exists, and will simply be
```Java
public Builder extraAnalyzeVisitor(Supplier<ClassVisitor> func) {
	extraAnalyzeVisitor = func;
	return this;
}

public Builder extraAnalyzeVisitor(ClassVisitor visitor) {
	return extraAnalyzeVisitor(() -> visitor);
}
```

LogicFan:
This should be a straightforward change. Fix the issue described in the #64

### DIFF
--- a/src/main/java/net/fabricmc/tinyremapper/TinyRemapper.java
+++ b/src/main/java/net/fabricmc/tinyremapper/TinyRemapper.java
@@ -157,6 +157,10 @@ public class TinyRemapper {
 		}
 
 		public Builder extraAnalyzeVisitor(ClassVisitor visitor) {
+			return this.extraAnalyzeVisitor(() -> visitor);
+		}
+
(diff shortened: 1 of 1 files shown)


sfPlayer1 (approved):

LogicFan:
Fixed in #63

PR #71, opened 2021-08-17
Try With Resources

CoolMineman:
I am tired of writing finally blocks. If you don't call remapper.finish the program won't exit because of the threadpool and this also makes that more obvious.


src/main/java/net/fabricmc/tinyremapper/TinyRemapper.java
```diff
@@ -291,7 +291,13 @@ public static Builder newRemapper() {
 		return new Builder();
 	}
 
+	@Deprecated
```
LogicFan, line 294: Please add Javadoc `@deprecated` to show users the new function to use.

PR #72, opened 2021-08-17
Separate API and implementation for tiny-remapper.

LogicFan:
WIP. Feedback is welcomed.


liach:
Imo we should just do a major version bump to like `0.6` so as to avoid having a lot of `Deprecated` stuff that's hard to clean up.

LogicFan:
Of course, that would be much easier to this kind of refactoring, but consider we just bumped the version to 0.5 not so long ago, I don't know if this is feasible. @sfPlayer1 what's your idea?

liach:
override autocloseable close to throw no checked exceptions so users can use try with resources without a catch block

LogicFan:
resolve #71 .

LogicFan:
Should have emulated all old API calls. Please let me know if I forget anything.

PR #73, opened 2021-08-24
Fix annotation being remapped twice

LogicFan:
This is a bug reported from downstream (Architectury). The annotation was being remapped twice. There is the fix https://github.com/architectury/tiny-remapper/commit/dc31879c2aaacb2cf274673b1f6ae22243a78ba3 from arch, but I think this PR has a more "correct" fix.


src/main/java/net/fabricmc/tinyremapper/AsmClassRemapper.java
```diff
@@ -197,23 +202,35 @@ public AnnotationVisitor visitTypeAnnotation(int typeRef, TypePath typePath, Str
...
-			return AsmClassRemapper.createAsmAnnotationRemapper(Type.getObjectType(owner).getDescriptor(), super.visitAnnotationDefault(), remapper);
+		public AnnotationVisitor visitAnnotationDefault() { // TODO: should we pass Type.getObjectType(owner).getDescriptor() or null for descriptor?
+			try {
+				AnnotationVisitor annotationVisitor = (AnnotationVisitor) MethodVisitor.class.getMethod("visitAnnotationDefault").invoke(this);
+				return annotationVisitor == null
+						? null
+						: createAnnotationRemapper(Type.getObjectType(owner).getDescriptor(), annotationVisitor);
+			} catch (Exception e) {
+				throw new RuntimeException("This should not happen!", e);
+			}
```
LogicFan, line 213: Why do we pass `Type.getObjectType(owner).getDescriptor()` instead of `null`. This confuses me considering the comment in 792 (https://github.com/LogicFan/tiny-remapper/blob/0dac35458491b163a1ba0d382a7675a93d3b3b70/src/main/java/net/fabricmc/tinyremapper/AsmClassRemapper.java#L792)

LogicFan:
I've re-write this PR with test-case. I think this one is correct, mind having a look @sfPlayer1 ?

sfPlayer1:
I don't think this is as good as some of my annotation remapping code. It (=asm's) doesn't always use the full desc where it could and doesn't implement enum field remapping. Please have a very close look at the differences between TR's AsmAnnotationRemapper and ASM's AnnotationRemapper

LogicFan:
I finished the rework for ASMAnnotationRemapper. Could you have a look again @sfPlayer1 , Thank you.

sfPlayer1 (approved):
LGTM, thanks!

PR #76, opened 2021-09-25
Allow custom executor in TinyRemapper

Devan-Kerman:


Devan-Kerman:
oops something crabbed git

PR #89, opened 2022-03-14
Fix #87

### LINKED ISSUE #87: Wrong remap due to static super method of same name
Given the following mappings, a remap of the local non-static method `getStringRaw` results in a mangled call due to the static super method with the same remapped name (being a synthetic lambda method):
```
c    net/minecraft/server/dedicated/Settings    net/minecraft/server/dedicated/PropertyManager
    m    (Ljava/lang/String;)Ljava/lang/String;    getStringRaw    c
c    net/minecraft/server/dedicated/DedicatedServerProperties    net/minecraft/server/dedicated/DedicatedServerProperties
    m    (Ljava/lang/String;)Ljava/lang/String;    lambda$new$3    c
```

The error is:
```
public class DedicatedServerProperties extends Settings<DedicatedServerProperties> { java.lang.IncompatibleClassChangeError: Expecting non-static method 'java.lang.String net.minecraft.server.dedicated.DedicatedServerProperties.c(java.lang.String)'
```

Specifically, this happens when cloning [Paper](https://github.com/PaperMC/Paper) and running the `runReobf` gradle task (or running the jar given by `createReobfBundlerJar`)
... **but apparently only when doing so on Windows**, not Mac or Linux

Any help or possible fix would be appreciated

KabanFriends:
This PR fixes the issue [Wrong remap of super method with local static method of same name #87](https://github.com/FabricMC/tiny-remapper/issues/87), so using tiny remapper on Windows should work properly.

Fix code provided by my friend who is unable to create a GitHub account.

### DIFF
--- a/src/main/java/net/fabricmc/tinyremapper/AsmClassRemapper.java
+++ b/src/main/java/net/fabricmc/tinyremapper/AsmClassRemapper.java
@@ -255,8 +255,9 @@ final class AsmClassRemapper extends VisitTrackingClassRemapper {
 			}
 
 			AsmRemapper asmRemapper = ((AsmRemapper) remapper);
+
 			for (ClassInstance checkClass = asmRemapper.getClass(owner); checkClass != null; checkClass = checkClass.getSuperClass()) {
-				if(checkClass.getMethod(name, descriptor) != null) {
+				if (checkClass.getMethod(name, descriptor) != null) {
 					owner = checkClass.getName();
 					break;
 				}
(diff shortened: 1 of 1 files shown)


KabanFriends:
Fixed, hopefully

sfPlayer1:
Thanks for the PR which helped outline the issue nicely, I however think this should be done a bit differently. I opened PR #90 since my approach is too different from yours. Please have a look and ideally a test if you can reproduce the original issue.

KabanFriends:
Sounds good, this PR can be closed at any time if the other one works better.

Lawrence-ux (approved):

PR #93, opened 2022-05-04
rework fs handler

sfPlayer1:
added a wrapper for easy closing (try-with-resource) and more convenience


src/main/java/net/fabricmc/tinyremapper/FileSystemReference.java
```diff
@@ -0,0 +1,150 @@
...
+			if (closed) return;
+			closed = true;
+
+			Set<FileSystemReference> refs = openFsMap.get(fileSystem);
+			if (refs == null || !refs.remove(this)) throw new IllegalStateException("fs "+fileSystem+" was already closed");
+
+			if (refs.isEmpty()) {
+				openFsMap.remove(fileSystem);
+				fileSystem.close();
+			} else if (refs.size() == 1 && refs.iterator().next() == null) { // only null -> not opened by us, just abandon
```
liach, line 132:
```suggestion
			} else if (refs.size() == 1 && refs.contains(null)) { // only null -> not opened by us, just abandon
```
Avoids allocation of iterator; and since identity hash map is a sparse hash table, iteration may be slower than hash lookup. See https://github.com/openjdk/jdk/blob/17cc7131abf699070dc491e69b21207f01c65cf6/src/java.base/share/classes/java/util/IdentityHashMap.java#L356

liach (approved):

PR #100, opened 2022-07-29
Avoid requiring identical fs in MetaInfFixer

TheGlitch76:
Uses the same workaround as in OutputConsumerPath


sfPlayer1 (approved):
I hate this limitation :D

PR #101, opened 2022-08-06
Only open JAR files if the provided input is a directory

TheGlitch76:
This PR removes some surprising behavior involving scanning JARs.

Previously, when presented with a JAR file, Tiny Remapper would open all of its nested JAR files and treat every file as an input. However, it would not re-construct the file structure. Instead, in the output it would flatten all nested JARs into the top jar, and create conflicts if the user had also added the nested JAR as an input.

This PR reverts that behavior, and makes Tiny Remapper ignore nested JARs.


sfPlayer1:
The behavior definitely need to change, not sure how exactly yet though. It'd theoretically be possible to (optionally) handle nested jars through sub-input-tags.

Issue #26, opened 2019-12-27
Scala invokedyamic bsm instructions are not handled


coderbot16:
* `unknown invokedynamic bsm: scala/runtime/LambdaDeserialize/bootstrap(Ljava/lang/invoke/MethodHandles$Lookup;Ljava/lang/String;Ljava/lang/invoke/MethodType;[Ljava/lang/invoke/MethodHandle;)Ljava/lang/invoke/CallSite; (tag=6 iif=false)`
* `unknown invokedynamic bsm: scala/runtime/SymbolLiteral/bootstrap(Ljava/lang/invoke/MethodHandles$Lookup;Ljava/lang/String;Ljava/lang/invoke/MethodType;Ljava/lang/String;)Ljava/lang/invoke/CallSite; (tag=6 iif=false)`

Example Forge mod jar with these `invokedynamic` instructions: https://www.curseforge.com/minecraft/mc-mods/additional-enchanted-miner/files/2848855

bluebear94:
I've encountered this problem, too, with [this mod](https://gitlab.com/Kyarei/hyphae-fabric).

lukebemish:
This problem also occurs for groovy Indy invokedynamic instructions:
```
unknown invokedynamic bsm: org/codehaus/groovy/vmplugin/v8/IndyInterface/bootstrap(Ljava/lang/invoke/MethodHandles$Lookup;Ljava/lang/String;Ljava/lang/invoke/MethodType;Ljava/lang/String;I)Ljava/lang/invoke/CallSite; (tag=6 iif=false)
```

It means I end up getting several hundred or thousand lines of log spam any time I build stuff. It's gotten to the point of making debugging difficult.

PR #106, opened 2023-03-19
Add CLI option to ignore invokedynamic types

lukebemish:
I've recently ran up against #26 in a project of mine - I get so much log spam due to tiny-remapper finding groovy indy instructions that it cannot remap that it is impossible to see any actual issues in the logs. I would like some way of turning off that warning for specific indy instructions, similar to what is already done for string concatenation. This PR adds a system property, `net.fabricmc.tinyremapper.knownindybsm` (this could definitely change, I'm by no means sold on it) for providing a list of instructions that tiny-remapper can safely assume do not need to be remapped.

If you think there may be a better approach to this than a system property, please let me know, and I can see about implementing that.

### DIFF
--- a/src/main/java/net/fabricmc/tinyremapper/AsmClassRemapper.java
+++ b/src/main/java/net/fabricmc/tinyremapper/AsmClassRemapper.java
@@ -26,8 +26,6 @@ import java.util.Objects;
 import java.util.Set;
 import java.util.function.Function;
 import java.util.regex.Pattern;
-import java.util.stream.Collectors;
-import java.util.stream.Stream;
 
 import javax.lang.model.SourceVersion;
 
--- a/src/main/java/net/fabricmc/tinyremapper/AsmClassRemapper.java
+++ b/src/main/java/net/fabricmc/tinyremapper/AsmClassRemapper.java
@@ -31,7 +31,6 @@ import java.util.stream.Stream;
 
 import javax.lang.model.SourceVersion;
 
-import net.fabricmc.tinyremapper.api.TrMember;
 import org.objectweb.asm.AnnotationVisitor;
 import org.objectweb.asm.ClassVisitor;
 import org.objectweb.asm.FieldVisitor;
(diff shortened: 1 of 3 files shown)


src/main/java/net/fabricmc/tinyremapper/AsmClassRemapper.java
```diff
@@ -277,12 +279,15 @@ public void visitInvokeDynamicInsn(String name, String descriptor, Handle bootst
 					bootstrapMethodArguments);
 		}
 
+		private static final Set<String> ALLOWED_INDY_BSM = Stream.of(System.getProperty("net.fabricmc.tinyremapper.knownindybsm", "").split(",")).collect(Collectors.toSet());
```
modmuss50, line 282: This should be set via the CLI and via the TinyRemapper.Builder. See how `forcePropagation` is set.

src/main/java/net/fabricmc/tinyremapper/AsmClassRemapper.java
```diff
@@ -277,12 +279,15 @@ public void visitInvokeDynamicInsn(String name, String descriptor, Handle bootst
...
+
 		private static Handle getLambdaImplementedMethod(String name, String desc, Handle bsm, Object... bsmArgs) {
 			if (isJavaLambdaMetafactory(bsm)) {
 				assert desc.endsWith(";");
 				return new Handle(Opcodes.H_INVOKEINTERFACE, desc.substring(desc.lastIndexOf(')') + 2, desc.length() - 1), name, ((Type) bsmArgs[0]).getDescriptor(), true);
 			} else if (bsm.getOwner().equals("java/lang/invoke/StringConcatFactory")
-					|| bsm.getOwner().equals("java/lang/runtime/ObjectMethods")) {
```
modmuss50, line 285: I think it would also be best to move these into the set.

modmuss50 (changes requested):

modmuss50:
Another option with this is to just not log issues at all, or maybe log it once (when a flag is set?)

lukebemish:
Alright, I *think* I hooked that up to the CLI system in a way that makes sense; let me know if not. I'm not quite sure enough about how TR works to try poking at adding the "warn only once" system, but that could always be added later as well.

modmuss50 (approved):

src/main/java/net/fabricmc/tinyremapper/Main.java
```diff
@@ -20,8 +20,11 @@
 
 import java.io.BufferedReader;
 import java.io.File;
+import java.io.FileInputStream;
 import java.io.FileReader;
```
modmuss50, line 24:
```suggestion
```
Removing this unused import.

PR #109, opened 2023-06-05
Implement barebone support for Mixin's `@Desc`

Geolykt:
Current drawbacks:
 - No "1:N" mappings (which may be the case for multiple owner classes)
 - No support for `@Desc` within `@At` (custom injection points would make it difficult to impossible to differ between method and field and thus it requires a different philosophy. However for my purposes I don't need `@Desc` in `@At` just yet)
 
 These drawbacks would require a novel idea to get resolved and in my opinion result from the fact that `@Desc` was not meant to get remapped.
 
 I believe I have briefly mentioned regex support within explicit target selectors, which in theory would allow to condense 1:N mappings down into a single regex string but even then we would get behavioral changes. Plus I don't know how far the capabilities of regex within explicit target selectors reach.
 
 
 As such the idea of this PR is more of a "it's better than nothing", with the sour taste of there being a few edge cases that were left unattended

### DIFF
--- a/src/main/java/net/fabricmc/tinyremapper/extension/mixin/soft/annotation/injection/CommonInjectionAnnotationVisitor.java
+++ b/src/main/java/net/fabricmc/tinyremapper/extension/mixin/soft/annotation/injection/CommonInjectionAnnotationVisitor.java
@@ -62,7 +62,7 @@ class CommonInjectionAnnotationVisitor extends FirstPassAnnotationVisitor {
 	public void visitEnd() {
 		// The second pass is needed regardless of remap, because it may have
 		// children annotation need to remap.
-		this.accept(new CommonInjectionSecondPassAnnotationVisitor(data, delegate, remap, targets, desc));
+		this.accept(new CommonInjectionSecondPassAnnotationVisitor(data, delegate, remap, targets));
 
 		super.visitEnd();
 	}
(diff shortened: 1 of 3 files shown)


src/main/java/net/fabricmc/tinyremapper/extension/mixin/soft/annotation/injection/CommonInjectionAnnotationVisitor.java
```diff
@@ -175,6 +178,24 @@ public void visit(String name, Object value) {
...
+			} else if (remap && name.equals(AnnotationElement.TARGET)) {	// All
+				return new AnnotationVisitor(Constant.ASM_VERSION, av) {
+					@Override
+					public AnnotationVisitor visitAnnotation(String name, String descriptor) {
+						if (!descriptor.equals(Annotation.DESC)) {
+							throw new RuntimeException("Unexpected annotation " + descriptor);
+						}
+
+						MemberType expectedDescType;
+						if (annotationDesc.equals(Annotation.MODIFY_CONSTANT)) {
```
modmuss50, line 190: This does not pass our checkstyle rules, causing the build to fail.

Geolykt: Updated. Completely forgot about the checkstyle before pushing

modmuss50 (approved):

modmuss50:
Im happy to merge this, and improve on it later if needed. The mixin support is still experimential 👍

PR #114, opened 2023-10-09
Support local variable remapping

NebelNidas:
Not sure why this wasn't enabled yet, from my (arguably limited) testing it looks like everything's working correctly.


SpaceWalkerRS:
I remember there being an issue with some obfuscated jars, since they lack a local variable table in the byte code. In that case this remapping would fail. I don't know if it's doable for TR to generate this byte code if any local variable mappings are given?

cph101:
This pull request is so simple, yet a total lifesaver. Thank you so much for implementing this! 🙏

modmuss50 (approved):

PR #138, opened 2024-10-15
Add a basic logger interface

modmuss50:
First stage of improving the log messages.

Has a slight breaking change to anyone who was passing a custom logger to the mixin extension.


modmuss50:
> Potential additions might be categories and context objects (like affected ClassInstance etc) but this should be fine for now.

Yes, being able to specify class/method context is the next thing I want to do. It will be especially useful for the mixin remapping.

PR #144, opened 2025-01-20
Feature/mixin improvements

LlamaLad7:
The first commit is to ignore Mixin `remap` values and just soft-fail if the name isn't remappable, like everywhere else. This is desirable for the user and simplifies the code quite a lot.
I tested this on the latest fabric-api and it produces a jar which is identical to that with `tiny-remapper#master`, as judged by running `jardiff` on all the nested jars.

However, the produced jar crashes due to some remapped accessor methods, hence my second commit. It is deeply undesirable to change the actual name of user-defined `Accessor`s and `Invoker`s, so instead if the name is meant to be inferred, we add the remapped name to the annotation. 

This is enough to build a fully working FAPI. Commits should be reviewed individually and not squashed.


modmuss50 (approved):
Much cleaner, thanks for this 👍

PR #147, opened 2025-06-19
fix: try resolving conflicts by using `getNewMappedName`

Norbiros:
Resolves https://github.com/architectury/architectury-loom/issues/283

We’re seeing this conflict error:

```
Mapping source name conflicts detected:
  kn METHOD c ((Lkp;)Z) -> [kn/method_57832, kn/method_57826]
    fixable: replaced with method_57832
  kl METHOD c ((Lkp;)Z) -> [kl/method_57826, kn/method_57832]
  km METHOD c ((Lkp;)Z) -> [kn/method_57832, km/method_57826]
    fixable: replaced with method_57826
  kn$a$a METHOD c ((Lkp;)Z) -> [kn/method_57832, kn$a$a/method_57826]
    fixable: replaced with method_57832
```

The only unfixable conflict is:

```
kl METHOD c ((Lkp;)Z) -> [kl/method_57826, kn/method_57832]
```

It should prefer `kl/method_57826`, but currently doesn't.

This PR adds an alternative conflict resolution strategy: if all existing checks fail (e.g. parent-based), we fall back to resolving by `member.cls.getName() + "/" + member.getNewMappedName()`.

This change is non-breaking and only runs as a last resort. It fixes the above issue without altering existing behaviour.

### DIFF
--- a/src/main/java/net/fabricmc/tinyremapper/TinyRemapper.java
+++ b/src/main/java/net/fabricmc/tinyremapper/TinyRemapper.java
@@ -894,7 +894,8 @@ public class TinyRemapper {
 					Map<String, String> mappings = member.type == TrMember.MemberType.METHOD ? methodMap : fieldMap;
 					String mappingName = mappings.get(member.cls.getName()+"/"+member.getId());
 
-					if (mappingName == null) { // no direct mapping match, try parents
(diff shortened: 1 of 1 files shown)


Norbiros:
This could be a possible fix, but this issue got already resolved in the architectury project

PR #160, opened 2026-02-02
Support method arg name remapping in `@ModifyVariable`

ishland:
This PR adds method arg name remapping in `@ModifyVariable`. This also adds a `TrLocal` API for such support.

Currently runtime remapping of fabric-api `0.143.0+26.1` to modern-yarn passes mixin audit after removing [method arg filtering in loader](https://github.com/FabricMC/fabric-loader/blob/088e1fe3487f64ae3b36eb0b3499def3167b944b/src/main/java/net/fabricmc/loader/impl/util/mappings/FilteringMappingVisitor.java#L34).

### DIFF
--- a/src/main/java/net/fabricmc/tinyremapper/TinyRemapper.java
+++ b/src/main/java/net/fabricmc/tinyremapper/TinyRemapper.java
@@ -57,6 +57,7 @@ import org.objectweb.asm.ClassReader;
 import org.objectweb.asm.ClassVisitor;
 import org.objectweb.asm.ClassWriter;
 import org.objectweb.asm.FieldVisitor;
+import org.objectweb.asm.Label;
 import org.objectweb.asm.MethodVisitor;
 import org.objectweb.asm.Opcodes;
 import org.objectweb.asm.RecordComponentVisitor;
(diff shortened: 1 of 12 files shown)


src/main/java/net/fabricmc/tinyremapper/TinyRemapper.java
```diff
@@ -662,7 +680,7 @@ public FieldVisitor visitField(int access, String name, String desc, String sign
 			cv = analyzeVisitors.get(i).insertAnalyzeVisitor(isInput, mrjVersion, name, cv, tags);
 		}
 
-		reader.accept(cv, ClassReader.SKIP_DEBUG | ClassReader.SKIP_FRAMES | ClassReader.SKIP_CODE);
+		reader.accept(cv, ClassReader.SKIP_FRAMES);
```
modmuss50: I wonder what the performance implications of this will be 🤔 If it is going to be an issue an option would be to make the locals opt-in for when they are needed.

modmuss50 (approved):

src/main/java/net/fabricmc/tinyremapper/TinyRemapper.java
modmuss50: Just thinking again, this will also increase the memory usage. I think having an option to either enable or disable tracking locals would be nice to have.

PR #164, opened 2026-04-26
Mixin method remap: fix desc being dropped & support wildcards

Moulberry:
### Fixed descriptor being dropped when input had descriptor

Descriptor should be used in output if and only if either:
1. Descriptor was present in input
2. Descriptor is necessary to disambiguate

### Added support for resolving multiple targets when using wildcard

See remapSeparateRemappedName test for example

### DIFF
--- a/src/main/java/net/fabricmc/tinyremapper/ClassInstance.java
+++ b/src/main/java/net/fabricmc/tinyremapper/ClassInstance.java
@@ -28,6 +28,7 @@ import java.util.Deque;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.IdentityHashMap;
+import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
@@ -907,7 +908,7 @@ public final class ClassInstance implements TrClass {
 	final Path srcPath;
 	byte[] data;
 	private ClassInstance mrjOrigin;
-	private final Map<String, MemberInstance> members = new HashMap<>(); // methods and fields are distinct due to their different desc separators
+	private final Map<String, MemberInstance> members = new LinkedHashMap<>(); // methods and fields are distinct due to their different desc separators
 	private final ConcurrentMap<String, MemberInstance> resolvedMembers = new ConcurrentHashMap<>();
 	final Set<ClassInstance> parents = new HashSet<>();
 	final Set<ClassInstance> children = new HashSet<>();
--- a/src/main/java/net/fabricmc/tinyremapper/extension/mixin/soft/annotation/injection/CommonInjectionAnnotationVisitor.java
+++ b/src/main/java/net/fabricmc/tinyremapper/extension/mixin/soft/annotation/injection/CommonInjectionAnnotationVisitor.java
@@ -108,8 +108,8 @@ class CommonInjectionAnnotationVisitor extends AnnotationVisitor {
 						throw new RuntimeException("InjectMethodMappable should never resolve to zero entries");
 					}
 
-					for (MemberInfo memberInfos : resolved) {
-						super.visit(name, memberInfos.toString());
+					for (MemberInfo remappedInfo : resolved) {
+						super.visit(name, remappedInfo.toString());
 					}
 				}
 			};
--- a/src/main/java/net/fabricmc/tinyremapper/extension/mixin/soft/annotation/injection/CommonInjectionAnnotationVisitor.java
+++ b/src/main/java/net/fabricmc/tinyremapper/extension/mixin/soft/annotation/injection/CommonInjectionAnnotationVisitor.java
@@ -104,6 +104,7 @@ class CommonInjectionAnnotationVisitor extends AnnotationVisitor {
 					}
 
 					List<MemberInfo> resolved = new InjectMethodMappable(data, info, targets).result();
+
 					if (resolved.isEmpty()) {
 						throw new RuntimeException("InjectMethodMappable should never resolve to zero entries");
 					}
@@ -183,6 +184,7 @@ class CommonInjectionAnnotationVisitor extends AnnotationVisitor {
 			desc = desc.isEmpty() ? null : desc;
 
 			Collection<TrMethod> col = owner.resolveMethods(name, desc, false, null, null);
+
 			if (col instanceof List) {
 				return (List<TrMethod>) col;
 			} else {
@@ -193,6 +195,7 @@ class CommonInjectionAnnotationVisitor extends AnnotationVisitor {
 		@Override
 		public List<MemberInfo> result() {
 			String mappedOwner = info.getOwner();
+
 			if (!mappedOwner.isEmpty()) {
 				mappedOwner = data.mapper.asTrRemapper().map(mappedOwner);
 			}
@@ -205,6 +208,7 @@ class CommonInjectionAnnotationVisitor extends AnnotationVisitor {
 				// Simple case when we can't find the specific method by name
 
 				String desc = info.getDesc();
+
 				if (!desc.isEmpty()) {
 					desc = data.mapper.asTrRemapper().mapDesc(desc);
 				}
@@ -221,6 +225,7 @@ class CommonInjectionAnnotationVisitor extends AnnotationVisitor {
 				List<TrMethod> methods = resolvePartials(target, info.getName(), info.getDesc());
 
 				int matchedCount = Math.min(methods.size(), methodsPerTarget);
+
 				for (int i = 0; i < matchedCount; i++) {
 					TrMember method = methods.get(i);
 
@@ -255,12 +260,15 @@ class CommonInjectionAnnotationVisitor extends AnnotationVisitor {
 					for (String mappedDesc : mappedDescriptors) {
 						if (canInject(mappedName, mappedDesc, fullMethodToTarget)) {
 							String quantifier = info.getQuantifier();
+
 							if (!explicitDesc) {
 								if (quantifierMin > 0) {
 									data.getLogger().error(Message.UNSUPPORTED_QUANTIFIER_MIN, info.toString());
 								}
+
 								quantifier = "";
 							}
+
 							list.add(new MemberInfo(mappedOwner, mappedName, quantifier, mappedDesc));
 						} else {
 							data.getLogger().error(Message.MISSING_INJECT, info.toString(), mappedName, mappedDesc);
@@ -286,6 +294,7 @@ class CommonInjectionAnnotationVisitor extends AnnotationVisitor {
 
 				for (TrMethod method : target.getMethods()) {
 					String otherName = data.mapper.mapName(method);
+
 					if (!otherName.equals(mappedName)) {
 						continue;
 					}
@@ -294,6 +303,7 @@ class CommonInjectionAnnotationVisitor extends AnnotationVisitor {
 
 					Pair<String, String> pair = Pair.of(otherName, otherDesc);
 					Set<TrClass> validClasses = fullMethodToTarget.get(pair);
+
 					if (validClasses == null || !validClasses.contains(target)) {
 						return false;
 					}
@@ -302,6 +312,7 @@ class CommonInjectionAnnotationVisitor extends AnnotationVisitor {
 					// e.g. If targeting method foo in [foo, bar -> baz, baz], we want to disambiguate the baz even though we would be
 					// targeting the correct method due to the max limit
 					toCheck -= 1;
+
 					if (toCheck <= 0 && methodsPerTarget > 1) {
 						break;
 					}
@@ -314,12 +325,14 @@ class CommonInjectionAnnotationVisitor extends AnnotationVisitor {
 		private boolean canInject(String mappedName, String mappedDesc, Map<Pair<String, String>, Set<TrClass>> fullMethodToTarget) {
 			Pair<String, String> pair = Pair.of(mappedName, mappedDesc);
 			Set<TrClass> validClasses = fullMethodToTarget.get(pair);
+
 			if (validClasses == null || validClasses.isEmpty()) {
 				return false;
 			}
 
 			for (TrClass target : targets) {
 				TrMethod method = target.getMethod(mappedName, mappedDesc);
+
 				if (method != null && !validClasses.contains(target)) {
 					return false;
 				}
@@ -337,18 +350,22 @@ class CommonInjectionAnnotationVisitor extends AnnotationVisitor {
 		if (quantifier == null || quantifier.isEmpty()) {
 			return Pair.of(0, 1);
 		}
+
 		if (quantifier.equals("*")) {
 			return Pair.of(0, Integer.MAX_VALUE);
 		}
+
 		if (quantifier.equals("+")) {
 			return Pair.of(1, Integer.MAX_VALUE);
 		}
+
 		if (!quantifier.startsWith("{") || !quantifier.endsWith("}") || quantifier.length() < 3) {
 			data.getLogger().error(Message.UNABLE_TO_PARSE_QUANTIFIER, quantifier);
 			return Pair.of(0, 0);
 		}
 
 		String inner = quantifier.substring(1, quantifier.length() - 1).trim();
+
 		if (inner.isEmpty()) {
 			data.getLogger().error(Message.UNABLE_TO_PARSE_QUANTIFIER, quantifier);
 			return Pair.of(0, 0);
@@ -358,6 +375,7 @@ class CommonInjectionAnnotationVisitor extends AnnotationVisitor {
 		String strMax = inner;
 
 		int comma = inner.indexOf(',');
+
 		if (comma > -1) {
 			strMin = inner.substring(0, comma).trim();
 			strMax = inner.substring(comma + 1).trim();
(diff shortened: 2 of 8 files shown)


src/main/java/net/fabricmc/tinyremapper/extension/mixin/soft/annotation/injection/CommonInjectionAnnotationVisitor.java
```diff
@@ -133,7 +152,7 @@ public AnnotationVisitor visitAnnotation(String name, String descriptor) {
 		return av;
 	}
 
-	private static class InjectMethodMappable implements IMappable<MemberInfo> {
+	private static class InjectMethodMappable implements IMappable<MemberInfo[]> {
```
LlamaLad7: Don't see any particular reason to use an array over a list.

src/main/java/net/fabricmc/tinyremapper/extension/mixin/soft/annotation/injection/CommonInjectionAnnotationVisitor.java
```diff
@@ -161,52 +180,144 @@ private Optional<TrMember> resolvePartial(TrClass owner, String name, String des
...
+			if (info.getDesc().isEmpty()) {
+				// If the descriptor was omitted in the input, we want to omit the descriptor in the output as well
+				// However, we can only do this if all the methods in the source namespace
+				// are exactly matched in the target namespace
+
+				for (Map.Entry<String, Set<String>> entry : descriptorsForName.entrySet()) {
+					String mappedName = entry.getKey();
+					Set<String> mappedDescriptors = entry.getValue();
+
+					Set<String> allDescriptorsInTargets = new HashSet<>();
```
LlamaLad7: Can simply be a count instead? Things match if the number of methods with that name is equal to the number we have

src/main/java/net/fabricmc/tinyremapper/extension/mixin/soft/annotation/injection/CommonInjectionAnnotationVisitor.java
```diff
@@ -161,52 +180,144 @@ private Optional<TrMember> resolvePartial(TrClass owner, String name, String des
...
+					}
 
-						return Pair.of(mappedName, shouldPassDesc ? data.mapper.mapDesc(m) : "");
-					})
-					.distinct().collect(Collectors.toList());
+					if (allDescriptorsInTargets.equals(mappedDescriptors)) {
+						finalMembers.add(new MemberInfo(data.mapper.asTrRemapper().map(info.getOwner()), mappedName, "*", ""));
+					} else {
+						for (String mappedDesc : mappedDescriptors) {
+							finalMembers.add(new MemberInfo(data.mapper.asTrRemapper().map(info.getOwner()), mappedName, "*", mappedDesc));
```
LlamaLad7: If we're passing a desc I probably wouldn't use a quantifier

src/main/java/net/fabricmc/tinyremapper/extension/mixin/soft/annotation/injection/CommonInjectionAnnotationVisitor.java
```diff
@@ -161,52 +180,144 @@ private Optional<TrMember> resolvePartial(TrClass owner, String name, String des
...
+						}
+					}
+				}
+			} else {
+				for (Map.Entry<String, Set<String>> entry : descriptorsForName.entrySet()) {
+					String mappedName = entry.getKey();
+					Set<String> mappedDescriptors = entry.getValue();
+
+					for (String mappedDesc : mappedDescriptors) {
+						finalMembers.add(new MemberInfo(data.mapper.asTrRemapper().map(info.getOwner()), mappedName, "*", mappedDesc));
```
LlamaLad7: Same here

src/main/java/net/fabricmc/tinyremapper/extension/mixin/soft/annotation/injection/CommonInjectionAnnotationVisitor.java
```diff
@@ -161,52 +180,144 @@ private Optional<TrMember> resolvePartial(TrClass owner, String name, String des
...
+					}
+				}
+			}
+
+			return finalMembers.toArray(new MemberInfo[0]);
+		}
+
+		private MemberInfo singleResult() {
+			if (targets.isEmpty() || info.getName().isEmpty()) {
+				return info;
```
LlamaLad7: What if we have a desc to remap but no name?

src/main/java/net/fabricmc/tinyremapper/extension/mixin/soft/annotation/injection/CommonInjectionAnnotationVisitor.java
```diff
@@ -161,52 +180,144 @@ private Optional<TrMember> resolvePartial(TrClass owner, String name, String des
...
+
+			if (collection.isEmpty()) {
+				data.getLogger().warn(Message.NO_MAPPING_NON_RECURSIVE, info.getName(), targets);
+				return new MemberInfo[] { info };
+			}
+
+			Map<String, Set<String>> descriptorsForName = new TreeMap<>();
+			for (Pair<String, String> pair : collection) {
+				descriptorsForName.computeIfAbsent(pair.first(), k -> new TreeSet<>()).add(pair.second());
+			}
```
LlamaLad7: Presumably the use of tree collections is for determinism? Could be good to mention or to type as `SortedMap` / `SortedSet` to make it explicit.

src/main/java/net/fabricmc/tinyremapper/extension/mixin/soft/annotation/injection/CommonInjectionAnnotationVisitor.java
```diff
@@ -85,11 +92,23 @@ public AnnotationVisitor visitArray(String name) {
...
+						return;
+					}
+
+					MemberInfo[] resolved = new InjectMethodMappable(data, info, targets).result();
+					if (resolved.length == 0) {
+						throw new RuntimeException("InjectMethodMappable should never resolve to zero entries");
+					}
+
+					for (MemberInfo memberInfos : resolved) {
+						super.visit(name, memberInfos.toString());
```
LlamaLad7: Nit: Why `memberInfos`?

src/main/java/net/fabricmc/tinyremapper/extension/mixin/soft/annotation/injection/CommonInjectionAnnotationVisitor.java
```diff
@@ -161,52 +180,144 @@ private Optional<TrMember> resolvePartial(TrClass owner, String name, String des
...
+		}
+
+		private MemberInfo[] wildcardResult() {
 			// Special case to remap the desc of wildcards without a name, such as `*()Lcom/example/ClassName;`
-			if (info.getOwner().isEmpty()
-					&& info.getName().isEmpty()
-					&& info.getQuantifier().equals("*")
-					&& !info.getDesc().isEmpty()) {
-				return new MemberInfo(info.getOwner(), info.getName(), info.getQuantifier(), data.mapper.asTrRemapper().mapDesc(info.getDesc()));
+			if (info.getName().isEmpty() && !info.getDesc().isEmpty()) {
```
LlamaLad7: Seems to have lost some of the conditions the previous impl had, notably `()I` should resolve to the first int method but after remapping it would now resolve to all of them.

LlamaLad7 (changes requested):

src/main/java/net/fabricmc/tinyremapper/extension/mixin/soft/annotation/injection/CommonInjectionAnnotationVisitor.java
Moulberry: If we're targeting multiple classes, we'd end up with a higher number.

class A {
   test()
}
class B {
   test()
   test(int)
}

We have 3 methods with the name test, but only 2 unique descriptors `test` and `test(int)`.
There might be other situations where using a count doesn't work, but this is the one that came to mind first

src/main/java/net/fabricmc/tinyremapper/extension/mixin/soft/annotation/injection/CommonInjectionAnnotationVisitor.java
Moulberry: [8e4e3b7](https://github.com/FabricMC/tiny-remapper/pull/164/commits/8e4e3b781b0f33b72f946264f80b7cdcb91d4814)

src/main/java/net/fabricmc/tinyremapper/extension/mixin/soft/annotation/injection/CommonInjectionAnnotationVisitor.java
Moulberry: Yes, for determinism.
[53ec0ec](https://github.com/FabricMC/tiny-remapper/pull/164/commits/53ec0ec7c0f0e6521e5eed97b787f8edf8f0b58d)

src/main/java/net/fabricmc/tinyremapper/extension/mixin/soft/annotation/injection/CommonInjectionAnnotationVisitor.java
Moulberry: [648fcf5](https://github.com/FabricMC/tiny-remapper/pull/164/commits/648fcf53aa952830b14bd665643fd72417d30ccd)

src/main/java/net/fabricmc/tinyremapper/extension/mixin/soft/annotation/injection/CommonInjectionAnnotationVisitor.java
LlamaLad7: I hadn't considered the case of multiple targets. But I think they need more special handling than that. We might match a given descriptor in only one of our targets for example.

That said it's easy to end up in unfixable situations generally with multi-target mixins, e.g. wanting to match `methodName()V` in one target but not another. I think the best thing we can do in such cases is error.

src/main/java/net/fabricmc/tinyremapper/extension/mixin/soft/annotation/injection/CommonInjectionAnnotationVisitor.java
Moulberry: [6677d5a](https://github.com/FabricMC/tiny-remapper/pull/164/commits/6677d5ad885d1388524653ca9880788ad298c3cd)

src/main/java/net/fabricmc/tinyremapper/extension/mixin/soft/annotation/injection/CommonInjectionAnnotationVisitor.java
Moulberry: [2e1a11b](https://github.com/FabricMC/tiny-remapper/pull/164/commits/2e1a11beec3d6c58f80bd956ad0bba51f7b70276)

src/main/java/net/fabricmc/tinyremapper/extension/mixin/soft/annotation/injection/CommonInjectionAnnotationVisitor.java
Moulberry: [2e1a11b](https://github.com/FabricMC/tiny-remapper/pull/164/commits/2e1a11beec3d6c58f80bd956ad0bba51f7b70276)

src/main/java/net/fabricmc/tinyremapper/extension/mixin/soft/annotation/injection/CommonInjectionAnnotationVisitor.java
Moulberry: [e151d28](https://github.com/FabricMC/tiny-remapper/pull/164/commits/e151d280c3c97602aae80a028b65efb80f9c42e3)

src/main/java/net/fabricmc/tinyremapper/extension/mixin/soft/annotation/injection/CommonInjectionAnnotationVisitor.java
Moulberry: Stale after rewrite

src/main/java/net/fabricmc/tinyremapper/ClassInstance.java
```diff
@@ -907,7 +908,7 @@ public static String getMrjName(String clsName, int mrjVersion) {
 	final Path srcPath;
 	byte[] data;
 	private ClassInstance mrjOrigin;
-	private final Map<String, MemberInstance> members = new HashMap<>(); // methods and fields are distinct due to their different desc separators
+	private final Map<String, MemberInstance> members = new LinkedHashMap<>(); // methods and fields are distinct due to their different desc separators
```
Moulberry, line 911: Should I sort based on TrMethod#getIndex or is making this a LinkedHashMap sufficient?

LlamaLad7: This is probably fine

src/main/java/net/fabricmc/tinyremapper/extension/mixin/soft/annotation/injection/CommonInjectionAnnotationVisitor.java
```diff
@@ -155,58 +176,174 @@ private static class InjectMethodMappable implements IMappable<MemberInfo> {
...
+				if (!wantDesc && canInject(mappedName, fullMethodToTarget)) { // Try to apply method name without descriptor if possible
+					list.add(new MemberInfo(mappedOwner, mappedName, info.getQuantifier(), ""));
+				} else {
+					for (String mappedDesc : mappedDescriptors) {
+						if (canInject(mappedName, mappedDesc, fullMethodToTarget)) {
+							String quantifier = info.getQuantifier();
+							if (quantifier.equals("*") || quantifier.startsWith("{0,")) {
+								quantifier = "";
 							}
+							list.add(new MemberInfo(mappedOwner, mappedName, quantifier, mappedDesc));
```
LlamaLad7, line 272: Can you elaborate on the goal behind these lines

LlamaLad7: Regarding quantifiers in this case, if the quantifier has a minimum of `1` or more, we cannot replicate this behaviour when emitting multiple name + desc pairs, so we should probably error

In all other cases we *should* just be able to keep the original quantifier

Moulberry: The quantifier is stripped to match the behaviour from this PR before the rewrite ([2e1a11b](https://github.com/FabricMC/tiny-remapper/pull/164/commits/2e1a11beec3d6c58f80bd956ad0bba51f7b70276))

> so if we know it's going to be * that we're inserting, we may as well just not

> If we're passing a desc I probably wouldn't use a quantifier

I don't even really know what the point of the minimum is inside mixins itself, so I don't even know what behaviour we would need to be replicating here

Okay yeah all it does is throw an error: <https://github.com/FabricMC/Mixin/blob/main/src/main/java/org/spongepowered/asm/mixin/injection/selectors/TargetSelectors.java#L248>

If there's a quantifier of 1 or more I'm not sure if it's better to show a warning, remove the quantifier and continue or error and not remap at all. I'd lean towards the first option

LlamaLad7: I could take or leave the quantifier stripping given we're now handling all kinds of quantifier, but if we're going to keep it presumably we want to handle `startsWith("{,")` too

Moulberry: True yes `startsWith("{,")` should also be included, technically { followed by any whitespace followed by , also works

Moulberry: Actually I just realized that, because of how the code works, using a max specifier like `method{0,4}` will always just try to replace it with the first 4 methods w/ descriptors. canInject() needs to be made smarter so it knows it can ignore methods if their match index is >4

Moulberry: I ended up just dropping the quantifier unless the input specified an explicit descriptor too (in which case we pass in the descriptor unchanged).
If the minimum quantifier is >1 we print a warning but still continue and drop the quantifier, because I think this is the best we can do in this case

desyd3780-cpu (approved):

desyd3780-cpu (approved):

modmuss50 (approved):

PR #166, opened 2026-05-19
Mixin regex remap in method selector

ishland:
Adds the capability to resolve mixin regex and remap resolved methods in method selectors.   
Passes mixin audit for fabric-api `0.149.1+26.2` with modern-yarn. 

The relevant mixin in fabric-api: https://github.com/FabricMC/fabric-api/blob/c3c10324a2b2973e6d1f9896a256f8df7f223c85/fabric-permission-api-v1/src/main/java/net/fabricmc/fabric/mixin/permission/CommandSourceStackMixin.java#L68

Resolving mixin regex in `@At.target` is not included for now and may come in a future PR. Pushing this out to get fabric-api working first.

### DIFF
--- a/src/main/java/net/fabricmc/tinyremapper/extension/mixin/soft/annotation/injection/CommonInjectionAnnotationVisitor.java
+++ b/src/main/java/net/fabricmc/tinyremapper/extension/mixin/soft/annotation/injection/CommonInjectionAnnotationVisitor.java
@@ -19,6 +19,7 @@
 package net.fabricmc.tinyremapper.extension.mixin.soft.annotation.injection;
 
 import java.util.Collections;
+import java.util.Comparator;
 import java.util.List;
 import java.util.Objects;
 import java.util.Optional;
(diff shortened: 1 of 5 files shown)


src/main/java/net/fabricmc/tinyremapper/extension/mixin/soft/annotation/injection/CommonInjectionAnnotationVisitor.java
```diff
@@ -85,7 +87,45 @@ public AnnotationVisitor visitArray(String name) {
...
+								.map(data.resolver::resolveClass)
+								.filter(Optional::isPresent)
+								.map(Optional::get)
+								.flatMap(trClass -> trClass.getMethods().stream())
+								.filter(trMethod -> matcher.matches(trMethod.getOwner().getName(), trMethod.getName(), trMethod.getDesc()))
+								.sorted(
+										Comparator
+												.<TrMethod, String>comparing(trMethod -> trMethod.getOwner().getName())
+												.thenComparing(TrMember::getName)
+												.thenComparing(TrMember::getDesc)
```
modmuss50: Indented 12 times, might be time to start thinking about creating a method.

ishland: Extracted it to `resolveAndRemapMixinRegex`

modmuss50 (approved):

